PR #32, opened 2022-06-15
fix: dyanmically determine dist version

agdimech:
The `test:integration` target of open-api-gateway assumes a dist version of 0.0.0. This is not always the case as for the `release:mainline` target, a bumped version is calculated.

This PR dynamically determines the version based on the presence of a version.txt file in dist.


cogwirrel (approved):

cogwirrel:
Thank you!

PR #33, opened 2022-06-15
fix(open-api-gateway): fix sample handler resolution, parsed spec import, and reduce generation time

cogwirrel:
A Couple of Fixes:
* Fix sample handler resolution by using the built in NodejsFunction's logic to look for ts then js files, so the handler can be found whether or not the api package is used as the cdk root
* Fix parsed spec import path when custom `specDir` or `parsedSpecFileName` are passed in

Codegen Performance Improvements:
* Prevent open-api-generator-cli from downloading open-api-generator every time in `generate` script by caching versions in `~/.open-api-generator-cli`
* Use pnpm (if installed) in the `generate` and `parse-openapi-spec` scripts for faster install time and therefore faster codegen time

Test speedup:
* Separate project tests into different files so jest can run them in parallel


packages/open-api-gateway/src/project/samples/typescript.ts
```diff
@@ -90,17 +97,15 @@ export class SampleApi extends Api {
       authType: AuthorizationType.IAM,
       integrations: {
         sayHello: {
-          function: new NodejsFunction(scope, "SayHelloHandler", {
-            entry: path.join(__dirname, "say-hello.handler.ts"),
-          }),
+          function: new NodejsFunction(scope, "say-hello"),
```
agdimech, line 100: no entry?

cogwirrel: If you don't specify an entry it infers it based on the id - this is where it checks for either `.ts` or `.js` which means this works whether it's a library in a monorepo (ie compiled to `.js`) or the main cdk target (ie not compiled).

agdimech: snazzy ;)

agdimech (approved):

PR #47, opened 2022-06-21
docs(nx-monorepo): add docs for homogenous dependencies

cogwirrel:


packages/nx-monorepo/README.md
```diff
@@ -94,5 +94,72 @@ The NxMonorepoProject also manages your yarn/pnpm workspaces for you and synthes
 
 For more information on NX commands, refer to this [documentation](https://nx.dev/using-nx/nx-cli).
 
+### Homogenous Dependencies
 
+As well as adding implicit dependencies, you can add dependencies between projects of the same language in order to have one project consume another project's code.
 
+#### Typescript
+
+Since the `NxMonorepoProject` manages a yarn/npm/pnpm workspace, configuring dependencies between Typescript projects is as straightforward as referencing them in deps:
```
agdimech: We should also mention that if trying to add deps between projects before projects are first synthed, you will run into an error as the project wont be present in the workspace yet and it will attempt to pull it down from NPM where it won't exist.

packages/nx-monorepo/README.md
```diff
@@ -94,5 +94,72 @@ The NxMonorepoProject also manages your yarn/pnpm workspaces for you and synthes
...
+  defaultReleaseBranch: "main",
+  name: "project-b",
+  // B depends on A
+  deps: [a.package.packageName],
+});
+```
+
+#### Python
+
+The recommended way to configure dependencies between python projects within your monorepo is to use a single shared virtual environment. You can then install packages you wish to depend on into that environment using pip's [Editable Installs](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs).
```
agdimech, line 149: Same here regarding synth order.

packages/nx-monorepo/README.md
cogwirrel: Good call!

packages/nx-monorepo/README.md
cogwirrel: Python actually works in one hit because the install to link two projects happens post-synth before anything is built :)

agdimech: Nice.

agdimech (approved):

PR #50, opened 2022-06-21
feat(open-api-gateway): add java client generation

cogwirrel:
Add java client generation.

Tested to ensure java client builds and can be depended upon within a monorepo.

Note that in order to configure the dependency, I added a local maven repository to the `pom.xml` of the dependant project, but this can't be done neatly via projen just yet (see https://github.com/projen/projen/issues/1943).

Once that issue is fixed, I'll add documentation on how the dependencies can be configured.


agdimech (approved):
Looks good 🚀

PR #52, opened 2022-06-21
fix(open-api-gateway): use projen .gitignore rather than open-api-generator

cogwirrel:
Since projen manages building the generated code, it knows better what should be ignored. Avoids
dist folders from being added to source control by default.


packages/open-api-gateway/src/project/codegen/components/open-api-generator-ignore-file.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Represents an .openapi-generator-ignore file
+ * @see https://github.com/OpenAPITools/openapi-generator/blob/master/docs/customization.md#ignore-file-format
+ */
+export class OpenApiGeneratorIgnoreFile extends IgnoreFile {
+  constructor(project: Project) {
+    super(project, ".openapi-generator-ignore");
+
+    // We should always use the projen .gitignore since projen manages the build, and therefore the ignored build
+    // artifacts
+    this.addPatterns(".gitignore");
```
agdimech, line 28: Doesn't this just add .gitignore as a line to this file?

cogwirrel: It does indeed! :) This tells openapi-generator-cli not to generate a `.gitignore` which overwrites the projen one :)

agdimech (approved):

PR #61, opened 2022-06-23
feat(open-api-gateway): python support

cogwirrel:
Add `OpenApiGatewayPythonProject` which has the same functionality as `OpenApiGatewayTsProject` but for python. This allows users to write their lambda handlers in python using the generated wrappers, and provides a type-safe cdk construct wrapper in python.


packages/open-api-gateway/README.md
```diff
@@ -50,6 +54,100 @@ If you would prefer to not generate the sample code, you can pass `sampleCode: f
...
+        |_ handlers/
+             |_ say_hello_handler_sample.py - An example lambda handler for the operation defined in spec.yaml, making use of the
+                                              generated lambda handler wrappers for marshalling and type safety.
+|_ generated/
+    |_ typescript/ - A generated typescript API client.
+    |_ python/ - A generated python API client, including generated lambda handler wrappers.
+    |_ java/ - A generated java API client.
+```
+
+For simplicity the generated code deploys a lambda layer for the generated code and its dependencies. You may choose to define an entirely separate projen `PythonProject` for your lambda handlers should you wish to add more dependencies than just the generated code.
```
agdimech: nitpick: Missing comma - For simplicity, the generated ....

packages/open-api-gateway/src/project/codegen/generated-python-client-project.ts
```diff
@@ -53,15 +58,21 @@ export class GeneratedPythonClientProject extends PythonProject {
...
-    ) {
+    if ((options.venv ?? true) && (options.pip ?? true)) {
       this.depsManager.installTask.exec("pip install --editable .");
     }
+
+    // Package into a directory that can be used as a lambda layer. This is done as part of install since the end user
+    // must control build order in the monorepo via explicit dependencies, and adding here means we can run as part of
+    // initial project synthesis which ensures this is created regardless of whether the user has remembered to
+    // configure build order
+    if (options.generateLayer) {
```
agdimech, line 71: :)

packages/open-api-gateway/src/project/open-api-gateway-python-project.ts
```diff
@@ -0,0 +1,230 @@
...
+ */
+export interface OpenApiGatewayPythonProjectOptions
+  extends PythonProjectOptions,
+    OpenApiGatewayProjectOptions {}
+
+/**
+ * Synthesizes a Python Project with an OpenAPI spec, generated clients, a CDK construct for deploying the API
+ * with API Gateway, and generated lambda handler wrappers for type-safe handling of requests.
+ *
+ * @pjid open-api-gateway-python
```
agdimech: I think we can just use `open-api-gateway-py` to be consistent.

agdimech (approved):

PR #73, opened 2022-07-01
feat: provide PDKNag util

agdimech:
- provide PDKNag util
- this only enables the util for a single package. Follow up PR to enable the rest and fix any nag related issues.

BREAKING CHANGE: rename type -> cidrType in CidrAllowList


packages/identity/cdk.out/Default.assets.json
```diff
@@ -0,0 +1,19 @@
+{
```
cogwirrel: Should we be ignoring `cdk.out`? :)

tools/pdk-nag/src/pdk-nag.ts
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * @inheritDoc
+ */
+export interface PDKNagAppProps extends AppProps {
+  /**
+   * Determines whether any errors encountered should trigger a test failure.
+   *
+   * @default true
+   */
+  readonly failOnError?: boolean;
```
cogwirrel, line 63: Might be nice to add `failOnWarning` for the extra-strict engineers among us :)

tools/pdk-nag/src/pdk-nag.ts
```diff
@@ -0,0 +1,142 @@
...
+      verbose: true,
+      reports: true,
+    });
+    this.app = app;
+  }
+
+  visit(node: IConstruct): void {
+    super.visit(node);
+
+    const results = node.node.metadata.filter(m => Object.values(CDK_NAG_MESSAGE_TYPES).find(v => v === m.type));
```
cogwirrel: Nit: could construct a global set of the message types and use `filter(m => CDK_NAG_MESSAGE_TYPES_SET.has(m.type))`

cogwirrel (dismissed):
Awesome! :)

packages/identity/cdk.out/Default.assets.json
agdimech: Ahh yep, that is an issue. Will remove.

tools/pdk-nag/src/pdk-nag.ts
agdimech: Not sure I understand this. Can you please elaborate?

cogwirrel: Just to reduce complexity from O(n*m) to O(n). m is 2 so not a big deal at all but it feels a bit neater to use a set rather than iterating the list of 2 items each time :)

```ts
const CDK_NAG_MESSAGE_TYPES = ...
const CDK_NAG_MESSAGE_TYPES_SET = new Set(Object.values(CDK_NAG_MESSAGE_TYPES));

...

const results = node.node.metadata.filter(m => CDK_NAG_MESSAGE_TYPES_SET.has(m.type));
```

cogwirrel (approved):

PR #85, opened 2022-07-12
fix: improve security posture of static-website

agdimech:


packages/static-website/src/cloudfront-web-acl.ts
```diff
@@ -93,53 +100,201 @@ export class CloudfrontWebAcl extends Construct {
...
+            new PolicyStatement({
+              effect: Effect.ALLOW,
+              actions: [
+                "logs:CreateLogGroup",
+                "logs:CreateLogStream",
+                "logs:PutLogEvents",
+              ],
+              resources: [
+                `arn:aws:logs:${stack.region}:${stack.account}:log-group:/aws/lambda/${onEventHandlerName}`,
+                `arn:aws:logs:${stack.region}:${stack.account}:log-group:/aws/lambda/${onEventHandlerName}:*`,
```
cogwirrel, line 140: Nit: Could use `stack.formatArn` here and throughout?

cogwirrel (approved):
Nice job! :)

packages/static-website/src/cloudfront-web-acl.ts
agdimech: Will make this change in a follow-up PR.

PR #86, opened 2022-07-12
fix(open-api-gateway): remove lib dir in precompile to ensure spec always updated

cogwirrel:
The parsed spec file was sometimes not being copied as part of build - it seems that
`tsc` doesn't always copy the file over if it's already there. Remove the lib as a
pre-compile step to force a fresh copy.

### DIFF
--- a/packages/open-api-gateway/src/project/open-api-gateway-ts-project.ts
+++ b/packages/open-api-gateway/src/project/open-api-gateway-ts-project.ts
@@ -131,6 +131,9 @@ export class OpenApiGatewayTsProject extends TypeScriptProject {
     });
     spec.synth();
 
+    // Delete the lib directory prior to compilation to ensure latest parsed spec json file is copied
+    this.preCompileTask.exec(`rm -rf ${this.libdir}`);
+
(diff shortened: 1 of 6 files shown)


packages/open-api-gateway/src/project/open-api-gateway-ts-project.ts
```diff
@@ -131,6 +131,9 @@ export class OpenApiGatewayTsProject extends TypeScriptProject {
     });
     spec.synth();
 
+    // Delete the lib directory prior to compilation to ensure latest parsed spec json file is copied
+    this.preCompileTask.exec(`rm -rf ${this.libdir}`);
```
agdimech, line 135: Looks like my suggestion worked ;)

agdimech (approved):

packages/open-api-gateway/src/project/open-api-gateway-ts-project.ts
cogwirrel: It did indeed! Much more robust than copying the single file 🎉  Made a test project and updated the spec.yaml a bunch of times and it always rebuilt :)

PR #87, opened 2022-07-12
fix: support nested stacks in CDK Nag suppressions

agdimech:
This PR adds nested stack support for Nag suppressions.

Previously we assumed these constructed were only apart of a Stack, however if contained within a NestedStack; errors would be encountered as the stack name was not complete.

This PR adds a util function called `getStackPrefix` which builds a complete Stack Prefix which can contain nested stack segments.

An example is say we have StackA that contains a nested StackB.

The resultant prefix should be `StackA/StackB/`.

### DIFF
--- a/packages/identity/src/nag-helper.ts
+++ b/packages/identity/src/nag-helper.ts
@@ -1,0 +1,33 @@
+/*********************************************************************************************************************
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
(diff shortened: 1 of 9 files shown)


agdimech:
I also opted to not use the `formatArn` function as per the previous PR as it was actually less readable.

cogwirrel:
> I also opted to not use the `formatArn` function as per the previous PR as it was actually less readable.

That's fair! :) Nice when it's a one liner :)

cogwirrel (approved):

PR #108, opened 2022-07-22
feat(open-api-gateway): support for chained request/response interceptors in typescript

cogwirrel:
Each generated handler wrapper now allows request interceptors to be passed as varargs. Interceptors are just functions which have the same interface as handler methods, but can delegate to the remainder of the chain to handle the request.

### DIFF
--- a/packages/open-api-gateway/README.md
+++ b/packages/open-api-gateway/README.md
@@ -1,6 +1,6 @@
 ## OpenAPI Gateway
 
-Define your APIs using [OpenAPI v3](https://swagger.io/specification/), and leverage the power of generated clients, automatic input validation, and type safe client and server code!
+Define your APIs using [OpenAPI v3](https://swagger.io/specification/), and leverage the power of generated clients and documentation, automatic input validation, and type safe client and server code!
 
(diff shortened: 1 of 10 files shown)


packages/open-api-gateway/README.md
```diff
@@ -433,6 +438,66 @@ def handler(input: SayHelloRequest, **kwargs) -> ApiResponse[HelloResponse, ApiE
...
+
+#### Typescript
+
+In typescript, interceptors are passed as separate arguments to the generated handler wrapper, in the order in which they should be executed.
+
+```ts
+import { sayHelloHandler, ApiError } from "my-api-typescript-client";
+
+// Interceptor to wrap invocations in a try/catch, returning a 500 error for any unhandled exceptions.
+const tryCatchInterceptor = async (input, event, context, chain) => {
```
agdimech: We dont have types for this interceptor however we do for timingInteceptor. Is this intended?

cogwirrel: Ah was just for brevity, but probably clearer if I include the types :)

agdimech (approved):
Looks good

PR #113, opened 2022-07-27
fix: corrects dependencies in sample projects

richardneililagan:
The pipeline Typescript samples refer to projen dependencies
(incorrectly) as just `aws-prototyping-sdk` prefixed, which doesn't
match the project declaration of the corresponding packages.

These have been corrected to reflect the right package name
declarations:

- `@aws-prototyping-sdk/pdk-nag`
- `@aws-prototyping-sdk/pipeline`

### DIFF
--- a/packages/pipeline/samples/typescript/src/pipeline-stack.ts
+++ b/packages/pipeline/samples/typescript/src/pipeline-stack.ts
@@ -13,8 +13,8 @@
  See the License for the specific language governing permissions and
  limitations under the License.
  ******************************************************************************************************************** */
+import { PDKPipeline } from '@aws-prototyping-sdk/pipeline';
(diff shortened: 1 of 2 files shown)


agdimech:
The existing dependencies are actually by design as these are classified as a STABLE  construct, hence present in aws-prototyping-sdk. This actually raises a bug as you will receive a failure if trying to consume this construct from the @aws-prototyping-sdk/pipeline directly.

richardneililagan:
That is interesting --- I ran into build issues from a fresh clone because of the items on this PR, hence the changes.

Could've been brought by something else though; let me give that another whirl and confirm.

agdimech:
Closing this PR.

PR #127, opened 2022-08-17
fix(cloudscape-react-ts-website): fix blank screen on signin due to missing runtime context

cogwirrel:
When first signing in and making use of runtime context, the context gets cleared, I think this is a race condition where the hub's signin callback fires before the useEffect runs to update the callback with the new runtime config. At any rate, using the setter method with the previous value addresses the issue! :)


agdimech (approved):

agdimech:
Good catch. Thankyou!

PR #138, opened 2022-08-26
feat(open-api-gateway): control re-generating clients and docs

sperka:
These changes make sure that re-generating client code and docs will happen only if it's necessary:
* spec changes;
* manually forced to generate (via parameter)
* new client language is added
* new documentationFormat is added

Changes:
* modify `ParsedSpec` to detect spec changes
* introduce `ClientSettings` to maintain client languages and document formats to generate
* introduce `forceGenerateCodeAndDocs` parameter for the project
* if not forced and no changes are made, projen won't re-generate the clients and documentation
* introduce `log4js` for logging (turn level to TRACE or DEBUG in the `rc` file to see logging in this package)
* chore: upgrade-deps

Fixes #


packages/open-api-gateway/src/index.ts
```diff
@@ -14,5 +14,10 @@
  limitations under the License.
  ******************************************************************************************************************** */
 
+// instantiate log4js when app is initialized
+import { join } from "path";
+import { configure as configureLog4js } from "log4js";
+configureLog4js(join(__dirname, "..", ".log4jsrc.json"));
```
cogwirrel: Will this file exist when imported as a library? I wonder if we need to make sure it's copied into the `lib` directory as part of the build - maybe even keep the file in `src` and import it and tsc will do it for you?

.npmrc
```diff
@@ -0,0 +1 @@
+yes=true
```
cogwirrel: We should probably have this managed by `.projenrc.ts` :)

packages/open-api-gateway/.log4jsrc.json
```diff
@@ -0,0 +1,14 @@
+{
```
cogwirrel: Same here - maybe best to generate this as part of `private/projects/open-api-gateway-project.ts`?

cogwirrel (commented):
Looks good! Thanks so much for this, will really speed up the dev cycle! :) Would be great to work on adding some tests to ensure it's generated once with no changes - hopefully something like this will do the trick:

* Instantiate and synth a new project
* Use fs.stat or similar to get the last modified time of a generated client file
* Instantiate and synth a second project with the same outdir as the previous
* Check that the last modified time has not changed

sperka:
@cogwirrel thanks for the feedback. will do and push the necessary changes soon

packages/open-api-gateway/src/index.ts
sperka: fixed with 45c75744f53f7dfb0ef905ecadbe5f70a4fdeede

.npmrc
sperka: fixed with #141

sperka:
added tests in 76c91a1ff7248ffc93a906889fb20c77a85d1f0b

Covers
* no spec change
* `forceGenerateCodeAndDocs = true` parameter
* spec change re-generates client and docs
* add new doc format doesn't trigger existing regeneration, but generates the newly added one

packages/open-api-gateway/src/project/open-api-gateway-ts-project.ts
```diff
@@ -150,28 +166,41 @@ export class OpenApiGatewayTsProject extends TypeScriptProject {
...
+    const clientSettings = new ClientSettings(this, {
+      clientLanguages: [...clientLanguages],
+      defaultClientLanguage: ClientLanguage.TYPESCRIPT,
+      documentationFormats: options.documentationFormats ?? [],
+      forceGenerateCodeAndDocs: this.forceGenerateCodeAndDocs,
+      generatedCodeDir: this.generatedCodeDir,
+      specChanged: spec.specChanged,
     });
 
+    // TODO: figure out testing
```
cogwirrel: Nit: you can get rid of this comment now :)

packages/open-api-gateway/test/project/open-api-gateway-ts-project/regenerate-clients.test.ts
```diff
@@ -0,0 +1,336 @@
...
+    );
+
+    // current `index.ts` should have later modified timestamp as the original project's
+    expect(generatedFileStats.mtime.getTime()).toBeLessThan(
+      generatedFileStatsOther.mtime.getTime()
+    );
+    // current `index.html` in "html2" doc folder should be untouched
+    expect(generatedDocHTML2Stats.mtime).toEqual(
+      generatedDocHTML2StatsOther.mtime
+    );
```
cogwirrel: Should the docs be regenerated too with `forceGenerateCodeAndDocs: true`?

packages/open-api-gateway/test/project/open-api-gateway-ts-project/regenerate-clients.test.ts
```diff
@@ -0,0 +1,336 @@
...
```
cogwirrel: Not for this pr but we should add some cleanup logic - the second project doesn't include any `documentationFormats` so ideally we would delete it :)

packages/open-api-gateway/test/project/open-api-gateway-ts-project/standalone.test.ts
```diff
@@ -34,5 +34,12 @@ describe("OpenAPI Gateway Ts Standalone Unit Tests", () => {
       packageManager,
     });
     expect(synthSnapshot(project)).toMatchSnapshot();
+
+    // testing that new invoke didn't happen
+    // TODO: check for existence of generated folder for TS/JAVA/PY
+    // then delete it
+    // create a new project2 with the same parameters --> no generated folder
+    // check for non-existense
+    // project.synth()
```
cogwirrel: Nit: can delete this :)

cogwirrel (commented):
Great stuff, couple of minor comments :)

packages/open-api-gateway/.log4jsrc.json
cogwirrel: No need for this since it's source code now :)

packages/open-api-gateway/src/project/open-api-gateway-ts-project.ts
sperka: 🙈 done

packages/open-api-gateway/src/project/client-config.ts
```diff
@@ -0,0 +1,27 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ ******************************************************************************************************************** */
+import { ClientLanguage, DocumentationFormat } from "./languages";
+
+export interface ClientLanguageConfig {
```
agdimech, line 21: We should add jsdocs for all these.

packages/open-api-gateway/src/project/codegen/components/client-settings.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  synthesize(): void {
+    // store the client config
+    // this helps to optimize to re-generate projects when new language/doc format is
+    // introduced while the spec doesn't change
+    logger.trace(`Generating settings file to ${this.clientSettingsPath}`);
+    const clientSettingsJsonFile = new JsonFile(
+      this.project,
+      this.clientSettingsPath,
```
agdimech, line 190: Do we need to add a gitignore for this file?

packages/open-api-gateway/test/project/open-api-gateway-ts-project/regenerate-clients.test.ts
sperka: the generator is called with `--minimal-update` switch, so actually the output is not changed, even though the generator runs. I've added a comment in the code describing this behavior

packages/open-api-gateway/test/project/open-api-gateway-ts-project/regenerate-clients.test.ts
sperka: agreed

packages/open-api-gateway/src/project/client-config.ts
sperka: done

packages/open-api-gateway/src/project/codegen/components/client-settings.ts
sperka: good call. added

cogwirrel (approved):
Awesome job! 🚀

agdimech (approved):
Nice work on your first PR, looks like it was a challenging one :)

PR #144, opened 2022-09-01
feat(open-api-gateway): add lambda handler wrappers and interceptors to generated java client

cogwirrel:
Add generation of Java lambda handler wrappers for each api operation.

Provides type-safe input and output for implementing api operations in Lambda, as well as interceptors for generic request/response handling to be shared across operations.

BREAKING CHANGE: Generated Java client methods now require `.execute()` to make the api call, eg
`api.sayHello("name").execute()`


agdimech (approved):
LGTM!

PR #148, opened 2022-09-02
fix(open-api-gateway): fix bug where all clients were generated no matter requested languages

cogwirrel:
Clients were always being generated no matter the requested clientLanguages - but only the empty
project shell without generated code! Addressed this by basing the clientSettigs on the input list
of languages rather than the whole enum. Additionally remove .client-settings.json from .gitignore
since generated client source code is checked in, so code generation does not need to run on first
checkout and build if the spec is unchanged.


packages/open-api-gateway/src/project/codegen/components/client-settings.ts
```diff
@@ -92,92 +92,50 @@ export class ClientSettings extends Component {
       options.generatedCodeDir,
       this.clientSettingsFilename
     );
-    // add client settings file to .gitignore
-    this.project.addGitIgnore(this.clientSettingsPath);
```
agdimech, line 96: Dont we still need this?

cogwirrel: No - I got rid of this because the generated clients are checked in, so if we also check this in we save a redundant regeneration of clients on first checkout and build :)

It's also useful to check this in so that we can implement cleanup in future - ie if you remove a generated client from the project definition we can see that it was previously there and delete the appropriate generated client directory :)

agdimech: OK, sgtm

agdimech (approved):

PR #152, opened 2022-09-07
feat(open-api-gateway): add java projen project type for defining cdk infrastructure in java

cogwirrel:
Add `OpenApiGatewayJavaProject` to complete the set of three (typescript, python and java)!

This generates the type-safe interface for the CDK construct in Java, as well as a sample construct and lambda handler for an out of the box hello world api.

Note also I address an issue with the generated python and java clients where nx was not picking up the generated client as a project in the repo, because the monorepo adds the `package.json` file but by that point the client had already been synthesized. Addressed this by adding the `package.json` ourselves and skipping it in the monorepo project. This enables us to set up the implicit dependency between the generated python client and the python api project (and the same for java) as well :).


packages/open-api-gateway/src/project/codegen/generated-java-client-project.ts
```diff
@@ -83,6 +84,19 @@ export class GeneratedJavaClientProject extends JavaProject {
...
+    new JsonFile(this, "package.json", {
+      obj: {
+        name: this.name,
+        __pdk__: true,
+        version: options.version,
+        scripts: Object.fromEntries(
+          this.tasks.all.map((task) => [task.name, `npx projen ${task.name}`])
+        ),
+      },
+      readonly: true,
```
agdimech, line 98: should we also set the marker given this is generated?

agdimech: same for all other langs.

packages/open-api-gateway/src/project/open-api-gateway-java-project.ts
```diff
@@ -0,0 +1,253 @@
...
+import { DocsProject } from "./codegen/docs-project";
+import { generateClientProjects } from "./codegen/generate";
+import { GeneratedJavaClientProject } from "./codegen/generated-java-client-project";
+import { ClientLanguage } from "./languages";
+import { getJavaSampleSource, JavaSampleCodeOptions } from "./samples/java";
+import { OpenApiSpecProject } from "./spec/open-api-spec-project";
+import { OpenApiGatewayProjectOptions } from "./types";
+
+const OPENAPI_GATEWAY_PDK_PACKAGE_NAME =
+  "software.aws.awsprototypingsdk/open-api-gateway@^0";
```
agdimech, line 30: well need to remember to bump this when we release 1.0

packages/open-api-gateway/src/project/open-api-gateway-java-project.ts
```diff
@@ -0,0 +1,253 @@
...
+      this.specFileName = "spec.yaml";
+    }
+    this.generatedCodeDir = options.generatedCodeDir ?? "generated";
+    this.forceGenerateCodeAndDocs = options.forceGenerateCodeAndDocs ?? false;
+    this.apiSrcDir = options.apiSrcDir ?? "api";
+
+    // Generated project should have a dependency on this project, in order to run the generation scripts
+    [
+      OPENAPI_GATEWAY_PDK_PACKAGE_NAME,
+      "software.constructs/constructs@^10",
```
agdimech, line 104: java deps in projen are painful...

agdimech (approved):

packages/open-api-gateway/src/project/open-api-gateway-java-project.ts
cogwirrel: Yeah it's a shame we have to specify a version :(

packages/open-api-gateway/src/project/open-api-gateway-java-project.ts
cogwirrel: Yeah, annoying!

packages/open-api-gateway/src/project/codegen/generated-java-client-project.ts
cogwirrel: I think given that all the rest of the generated client code won't have the marker there's not much benefit in adding it for the `package.json`. Greg is actually looking into adding more general cleanup of client projects - ie when you remove a `ClientLanguage` it deletes the appropriate directory, or when you change the spec you don't end up with old dangling generated files :)

agdimech: Ok sounds good.

PR #153, opened 2022-09-07
feat(open-api-gateway): add aws wafv2 webacl to api, enabled by default

cogwirrel:
Adds a WebACL by default, with options to customise its ruleset or completely disable it.


packages/open-api-gateway/src/construct/open-api-gateway-lambda-api.ts
```diff
@@ -331,6 +346,16 @@ export class OpenApiGatewayLambdaApi extends Construct {
       });
     });
 
+    // Create and associate the web acl
+    const acl = new OpenApiGatewayWebAcl(this, `${id}-Acl`, {
```
agdimech: Shouldn't we should only create this if `!props.webAclOptions.disable`?

I know you have it within the construct itself but it doesn't feel like the right place to me.

cogwirrel: Sure I can move it out :)

agdimech (approved):

PR #159, opened 2022-09-12
fix: upgrade version of python open api generator

em-xqm:
Fixes #

upgrade openapi tool's python generator from version v6.0.0 to v6.1.0


packages/open-api-gateway/test/project/open-api-gateway-java-project/__snapshots__/standalone.test.ts.snap
```diff
@@ -7283,7 +7365,7 @@ RecursionError indicating the maximum recursion limit has been exceeded. In that
 
 Solution 1:
 Use specific imports for apis and models like:
-- \`from myapi_python.api.default_api import DefaultApi\`
+- \`from myapi_python.apis.default_api import DefaultApi\`
```
cogwirrel, line 7368: I think we should mark this as a breaking change since the import path for the api client has changed so this will break api client consumers :) Doesn't look like anything the generated handlers etc relies on has changed which is good :) Please could you run `npx cz` again and hit `y` for a breaking change? :)

cogwirrel (changes requested):

cogwirrel (dismissed):

agdimech (dismissed):

cogwirrel (approved):

PR #160, opened 2022-09-13
feat(open-api-gateway): support smithy as alternative api interface definition language

### LINKED ISSUE #145: [FEATURE] [open-api-gateway] Smithy Support
### Describe the feature

As an alternative to OpenAPI, it would be great if APIs could be defined with [Smithy](https://awslabs.github.io/smithy/2.0/index.html), a protocol-agnostic IDL.

### Use Case

Defining APIs in an alternative, non-yaml format, less verbose than OpenAPI without the need to understand `$ref` and `allOf` etc.

### Proposed Solution

New projen constructs `SmithyApiGateway<Language>Project` could be exposed by the `open-api-gateway` package which will generate the sample hello world api in Smithy IDL.

Smithy IDL can be converted to OpenAPI, so we would add this as a pre-step such that the generated clients, construct etc can be reused.

Smithy IDL allows for authorizers to be specified, so we would need to figure out a neat way to marry up those specified in the `OpenApiGatewayLambdaApi` CDK construct with those specified in the spec.

### Other Information

Generated lambda handler wrappers would all still be generated and in theory would work with no change, however we might also wish to provide the option to use the [Smithy Server SDK for Typescript](https://awslabs.github.io/smithy/2.0/ts-ssdk/index.html) as an alternative for writing type-safe lambda handlers in typescript.

### Acknowledgements

- [X] This feature might incur a breaking change

### PDK version used

0.8.x

Typescript, Java, Python

All!

cogwirrel:
Adds three new projen project types (ts/python/java) where Smithy can be used instead of OpenAPI as the interface definition language. Smithy is projected into OpenAPI as a pre-step, which allows us to reuse the same generated clients, handler wrappers and infrastructure.

Fixes #145


agdimech (commented):
Do we need to check in the gradle-wrapper.jar?

packages/open-api-gateway/src/construct/prepare-spec-event-handler/prepare-spec.ts
```diff
@@ -279,6 +279,140 @@ const preparePathSpec = (
...
 
+/**
+ * Return whether the given OpenAPI object is a reference object
+ */
+const isRef = (obj: any): obj is OpenAPIV3.ReferenceObject => "$ref" in obj;
+
+/**
+ * Validate the construct security schemes against the security schemes in the original spec.
+ * Construct-defined authorizers always override those in the spec if they have the same ID, however we validate that
+ * we re not overriding an authorizer of a different type to avoid mistakes/mismatches between the spec and the
```
agdimech: typo: we are* not

cogwirrel:
> Do we need to check in the gradle-wrapper.jar?

We do to avoid having a `gradle` install being a prerequisite :) We use the one I checked in to run the smithy build unless it's overridden in the `gradleWrapperPath` property :)

packages/open-api-gateway/src/project/smithy-api-gateway-python-project.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export interface SmithyApiGatewayPythonProjectOptions
+  extends PythonProjectOptions,
+    SmithyApiGatewayProjectOptions {}
+
+/**
+ * Synthesizes a Python Project with a Smithy model, generated clients, a CDK construct for deploying the API
+ * with API Gateway, and generated lambda handler wrappers for type-safe handling of requests.
+ *
+ * @pjid smithy-api-gateway-python
```
agdimech: pjid should end with -py to be consistent with the open api construct.

packages/open-api-gateway/src/project/open-api-gateway-ts-project.ts
```diff
@@ -98,20 +98,22 @@ export class OpenApiGatewayTsProject extends TypeScriptProject {
...
-          ...options.tsconfig?.compilerOptions,
+          ...projectOptions.tsconfig?.compilerOptions,
         },
       },
     });
 
-    if (options.specFile) {
+    const options = this.preConstruct(projectOptions);
+
+    if (options.specFile && !path.isAbsolute(options.specFile)) {
```
agdimech, line 116: is the specFile is not absolute then we ignore the value and revert to spec.yaml?

agdimech: Is this documented?

packages/open-api-gateway/src/project/smithy-api-gateway-python-project.ts
cogwirrel: ah good catch, thanks!

packages/open-api-gateway/src/project/open-api-gateway-ts-project.ts
cogwirrel: So if it's not absolute we'll still generate a sample hello world spec in that location relative to the project source dir, eg `specFile: 'my-folder/my-spec.yaml'` will result in the hello world spec being written to `src/my-folder/my-spec.yaml`.

Then if it's absolute we skip generating the sample and just use that as the input spec to parse :)

The docs say to use a path relative to the project source dir - using an absolute path directly would mean that users can't really share their code so I didn't want to document it as an available feature :)

agdimech: Makes sense.

agdimech (approved):

PR #163, opened 2022-09-15
feat(open-api-gateway): support integrations other than lambda

cogwirrel:
Rather than enforce that an integration is an `IFunction`, instead we accept an `Integration` base class. Integrations must implement a `render` method which must return the `x-amazon-apigateway-integration` snippet to add to the prepared spec.

For now, we still only provide a `LambdaIntegration` but more can be added in the future, and users can extend the `Integration` base class to implement their own as an escape hatch. Also rename the construct from `OpenApiGatewayLambdaApi -> OpenApiGatewayRestApi` since we're no longer tied to lambda.

This change also scopes down integration permissions from API Gateway -> Lambda to the specific stage/method/path of the operation.

Additionally, remove pytest from the Python api project since it doesn't come with tests! Instead allow users to enable pytest if they like.

BREAKING CHANGE: Rename `OpenApiGatewayLambdaApi` construct to `OpenApiGatewayRestApi`. Integrations no longer take a `function` prop, and instead must be given `integration: Integrations.lambda(myLambda)`


packages/open-api-gateway/README.md
```diff
@@ -618,6 +618,20 @@ Authorizers.custom({
...
+
+##### Lambda Integration
+
+Currently, the only built-in integration is a lambda integration. You can construct one using `Integrations.lambda(yourLambdaFunction)`.
+
+##### Custom Integrations
+
+You can implement your own integrations by inheriting the `Integration` class and implementing its `render` method. This method is responsible for returning a snippet of OpenAPI which will be added as the `x-amazon-apigateway-integration` for an operation. Please refer to the [API Gateway Swagger Extensions documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-integration.html) for more details.
+
+You can also optionally override the `grant` method if you need to use CDK to grant API Gateway access to invoke your integration.
```
agdimech, line 633: Wouldn't you always want to do this?

packages/open-api-gateway/src/construct/integrations/integrations.ts
```diff
@@ -0,0 +1,284 @@
...
+   */
+  public grant({ scope, api, operationId }: IntegrationGrantProps) {
+    new CfnPermission(scope, `LambdaPermission-${operationId}`, {
+      action: "lambda:InvokeFunction",
+      principal: "apigateway.amazonaws.com",
+      functionName: this.lambdaFunction.functionArn,
+      sourceArn: Stack.of(scope).formatArn({
+        service: "execute-api",
+        resource: api.restApiId,
+        resourceName: "*/*",
```
agdimech: given one integration is servicing a single route, shouldn't this be specific to said route?

packages/open-api-gateway/README.md
cogwirrel: It depends - you might be proxying to some third party non-aws http endpoint in which case there's no iam permissions etc to set up :)

packages/open-api-gateway/src/construct/integrations/integrations.ts
cogwirrel: Yes it's likely possible to scope down - would require a bit of extra work to translate the method+path to a resource (eg we'd need to substitute path parameters with `*`) so might be best saved for a future enhancement? :)

packages/open-api-gateway/README.md
agdimech: Makes sense.

packages/open-api-gateway/src/construct/integrations/integrations.ts
cogwirrel: Updated to scope down to the specific stage/method/path :)

agdimech (approved):

PR #166, opened 2022-09-19
feat(open-api-gateway): prefer user-specified dependency versions for auto-added dependencies

### LINKED ISSUE #164: [FEATURE] Be able to lock the version of python open-api-gateway api project
### Describe the feature

Currently, there is no way to lock the version of python *@aws-prototyping-sdk/open-api-gateway* deps in the API project itself since the project is automatically generated. 

Even when the version of *@aws-prototyping-sdk/open-api-gateway* is locked in the monorepo definition, the version in the API project is not locked. 

When there is any breaking change in the lib, the CI/CD pipeline will break since it is always installing the latest version if we cannot lock the version from requirements.txt or use tooling like pipenv. 

### Use Case

Now when there is any breaking change in the lib, the CI/CD pipeline will break since it is always installing the latest version if we cannot lock the version from requirements.txt or use tooling like pipenv. 

### Proposed Solution

The *@aws-prototyping-sdk/open-api-gateway* should have the same version as the monorepo project if it is locked so that we have consistent build result. 

### Other Information

This is when I tried to lock the version: 

projen.rc file
```
const monorepo = new nx_monorepo.NxMonorepoProject({
  defaultReleaseBranch: 'mainline',
  devDeps: ['aws-prototyping-sdk@0.10.2'],
  deps: [
    '@aws-prototyping-sdk/cloudscape-react-ts-website@0.10.2', 
    '@aws-prototyping-sdk/open-api-gateway@0.10.2'
  ],
  gitignore: [".env", ".ipynb_checkpoints"],
  name: '******',
  tsconfig: {
    compilerOptions: {},
    include: [
      "packages/**/*.ts",
      "packages/**/*.tsx",
    ]
  }
});
```

deps.json file of the monorepo 
```
```
    {
      "name": "@aws-prototyping-sdk/open-api-gateway",
      "version": "0.10.2",
      "type": "runtime"
    },
```

deps.json file of the api project
```
    {
      "name": "aws_prototyping_sdk.open_api_gateway",
      "type": "runtime"
    },
```



### Acknowledgements

### PDK version used

0.11.0

Python

CodePipeline

### PR DESCRIPTION
**Pull request by cogwirrel**
Where a user has specified a dependency version, we prefer that over the ones we add automatically. This allows for pinning of the `@aws-prototyping-sdk/open-api-gateway` package version, as well as cdk, constructs etc.

Fixes #164

### DIFF
--- a/packages/open-api-gateway/src/project/open-api-gateway-python-project.ts
+++ b/packages/open-api-gateway/src/project/open-api-gateway-python-project.ts
@@ -15,7 +15,13 @@
  ******************************************************************************************************************** */
 
 import * as path from "path";
-import { Project, SampleDir, SampleFile, TextFile } from "projen";
+import {
+  DependencyType,
+  Project,
+  SampleDir,
+  SampleFile,
+  TextFile,
+} from "projen";
 import { PythonProject, PythonProjectOptions } from "projen/lib/python";
 import { ClientSettings } from "./codegen/components/client-settings";
 import { DocsProject } from "./codegen/docs-project";
@@ -112,12 +118,9 @@ export class OpenApiGatewayPythonProject extends PythonProject {
     this.apiSrcDir = options.apiSrcDir ?? "api";
 
     // Generated project should have a dependency on this project, in order to run the generation scripts
-    [
-      OPENAPI_GATEWAY_PDK_PACKAGE_NAME,
-      "constructs",
-      "aws-cdk-lib",
-      "cdk-nag",
-    ].forEach((dep) => this.addDependency(dep));
+    [OPENAPI_GATEWAY_PDK_PACKAGE_NAME, "constructs", "aws-cdk-lib", "cdk-nag"]
+      .filter((dep) => !this.deps.tryGetDependency(dep, DependencyType.RUNTIME))
+      .forEach((dep) => this.addDependency(dep));
 
     // Synthesize the openapi spec early since it's used by the generated python client, which is also synth'd early
     const spec = new OpenApiSpecProject({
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: packages/open-api-gateway/src/project/open-api-gateway-python-project.ts [hunk 1]
```diff
@@ -112,12 +118,9 @@ export class OpenApiGatewayPythonProject extends PythonProject {
...
 
     // Generated project should have a dependency on this project, in order to run the generation scripts
-    [
-      OPENAPI_GATEWAY_PDK_PACKAGE_NAME,
-      "constructs",
-      "aws-cdk-lib",
-      "cdk-nag",
-    ].forEach((dep) => this.addDependency(dep));
+    [OPENAPI_GATEWAY_PDK_PACKAGE_NAME, "constructs", "aws-cdk-lib", "cdk-nag"]
+      .filter((dep) => !this.deps.tryGetDependency(dep, DependencyType.RUNTIME))
```
agdimech, line 122: What about pinned versions? Do we need to split the dep on @ here as well?

packages/open-api-gateway/src/project/open-api-gateway-ts-project.ts
```diff
@@ -138,10 +144,14 @@ export class OpenApiGatewayTsProject extends TypeScriptProject {
...
-      "constructs",
-      "aws-cdk-lib",
-      "cdk-nag"
+      ...[
+        OPENAPI_GATEWAY_PDK_PACKAGE_NAME,
+        "constructs",
+        "aws-cdk-lib",
+        "cdk-nag",
+      ].filter(
+        (dep) => !this.deps.tryGetDependency(dep, DependencyType.RUNTIME)
```
agdimech, line 153: same here.

packages/open-api-gateway/src/project/open-api-gateway-python-project.ts
cogwirrel: The list of versions we're filtering doesn't have any `@`s in it so I omitted the splitting :)

agdimech (approved):

PR #180, opened 2022-09-30
fix: Enable aws api gateway test console to invoke lambdas

### LINKED ISSUE #178: [BUG] Cannot use api test console when using lambda integration
### Describe the bug

Since https://github.com/aws/aws-prototyping-sdk/blob/mainline/packages/open-api-gateway/src/construct/integrations/integrations.ts#L260 grants access only to the deployment stage, the api gateway test console (which uses a stage named "test-invoke-stage") fails invoking any lambda with an error `Execution failed due to configuration error: Invalid permissions on Lambda function`

I suggest either adding this stage to the lambda resource policy generated, or using a "*" for stage name in the same way api gateway does when associating a lambda with a method

### Expected Behavior

I can invoke a lambda integration through the AWS Api gateway test console

### Current Behavior

The lambda fails with "Invalid permissions on Lambda function"

### Reproduction Steps

Generate an api gateway with a lambda integration using the @aws-prototyping-sdk/open-api-gateway module

### Possible Solution

I submitted a [PR](https://github.com/aws/aws-prototyping-sdk/pull/180) changing from the specific stage to a star permission for the lambda integration

_No response_

### PDK version used

0.11.4

_No response_

Any

AndreaCatalucciTractable:
Fixes #178

I propose using a "*" as stage name, I do not see any immediate security threats and I believe the general intent of lambda integrations is for them to be enabled for every stage available on the api gateway.
This follows the pattern of the resource policy which is auto-generated from the aws console when adding a new lambda integration to a method

### DIFF
--- a/packages/open-api-gateway/src/construct/integrations/integrations.ts
+++ b/packages/open-api-gateway/src/construct/integrations/integrations.ts
@@ -271,11 +271,12 @@ export class LambdaIntegration extends Integration {
       sourceArn: Stack.of(scope).formatArn({
         service: "execute-api",
         resource: api.restApiId,
-        // Scope permissions to the stage, method and path of the operation.
+        // Scope permissions to any stage and a specific method and path of the operation.
         // Path parameters (eg {param} are replaced with wildcards)
-        resourceName: `${
-          api.deploymentStage.stageName
-        }/${method.toUpperCase()}${path.replace(/{[^\}]*\}/g, "*")}`,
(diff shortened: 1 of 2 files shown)


agdimech (approved):
Thanks for creating this PR! Nice job ;)

cogwirrel (approved):
Sounds great - it's still restricted to the specific api, method and path so I don't see any security concerns here. Thanks for the PR!

Issue #177, opened 2022-09-29
[FEATURE] Smithy generated Server SDK (SSDK) for TS


codypenta:
### Describe the feature

It looks like we generate types and values from OpenAPI tools. It would be nice to generate types, validators, etc using Smithy tools. Right now, it looks like most are unstable with the exception of typescript, which is in preview:

- https://aws.amazon.com/blogs/devops/smithy-server-and-client-generator-for-typescript/
- https://awslabs.github.io/smithy/2.0/ts-ssdk/index.html
- https://github.com/aws-samples/smithy-server-generator-typescript-sample

### Use Case

From previous experiences, the code generation tends to be better coming from a Smithy Model vs going from Smithy -> OpenAPI -> OpenAPITools but not in any *drastic* way.

### Proposed Solution

By default, a SmithyApiGatewayTsProject would use smithy code generation, but you can opt-in to OpenAPI generation

```ts
const api = new SmithyApiGatewayTsProject({
  clientLanguages: [ClientLanguage.TYPESCRIPT],
  defaultReleaseBranch: "main",
  name: "api",
  parent: monorepo,
  useOpenAPIGen: true, // <-- new flag
  outdir: "packages/api",
  serviceName: "aws.bumpercars#Bumpercars",
  deps: [
    "@aws-lambda-powertools/logger",
    "@aws-lambda-powertools/tracer",
    "@aws-lambda-powertools/metrics",
  ],
});
```

### Acknowledgements

- [X] I may be able to implement this feature request
- [X] This feature might incur a breaking change

### PDK version used

0.11.3

### What languages will this feature affect?

Typescript

### Environment details (OS name and version, etc.)

MacOS

codypenta:
On this, it looks like the Smithy stuff is exported as well so I was able to make a `packages/model` and `packages/codegen`


```typescript
new SmithyModelProject({
  name: "model",
  outdir: "packages/model",
  serviceName: "Bumpercars",
  serviceNamespace: "aws.bumpercars",
  parent: monorepo
})

new SmithyBuildProject({
  buildOutputDir: "codegen",
  fullyQualifiedServiceName: "aws.bumpercars#Bumpercars",
  modelPath: "../model/main.smithy",
  name: "codegen",
  parent: monorepo,
  outdir: "packages/codegen",
  smithyBuildOptions: {
    plugins: {
      "typescript-codegen": {
        "service": "aws.bumpercars#Bumpercars",
        "targetNamespace": "Configuration",
[...]

PR #183, opened 2022-10-04
feat(nx-monorepo): support adding custom additional packages to the yarn/npm/pnpm workspace

cogwirrel:
Allows adding additional packages to the yarn/npm/pnpm workspace via props or the `addWorkspacePackages` method. These are added in addition to those automatically added by parenting them with the monorepo in the `.projenrc`. This allows you to add packages which may not be controlled by projen, for example code generated by other tools.

re #177


packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -337,6 +347,14 @@ export class NxMonorepoProject extends TypeScriptProject {
     }
   }
 
+  /**
+   * Add an additional package glob to the workspace.
+   * @param packageGlob path to the package to include in the workspace (for example packages/my-package)
+   */
+  public addWorkspacePackage(packageGlob: string) {
```
agdimech: Can we support vararg so multiple can be added in one call?

agdimech (dismissed):

packages/nx-monorepo/src/nx-monorepo.ts
cogwirrel: Yeah good idea!

agdimech (approved):

PR #184, opened 2022-10-05
feat(open-api-gateway): make service name a required property for smithy projects

### LINKED ISSUE #176: [BUG] Smithy Service name must be defined ahead of time
### Describe the bug

When you generate a sample project, then manually overwrite the smithy sample generated. `npx projen` will not rebuild the smithy model.

### Expected Behavior

Able to overwrite the generated smithy model file and rebuild to get new types.

### Current Behavior

`npx projen` will not recognize input change and will not build out new model

### Reproduction Steps

1.) generate smithy open api
2.) manually change namespace, service, and operations
3.) Run `npx projen`

### Possible Solution

This was fixed by deleted the generated API, then typing the service name ahead of time:

```ts
const api = new SmithyApiGatewayTsProject({
  clientLanguages: [ClientLanguage.TYPESCRIPT],
  defaultReleaseBranch: "main",
  name: "api",
  parent: monorepo,
  outdir: "packages/api",
  serviceName: "aws.bumpercars#Bumpercars", <-- add this, then npx projen
  deps: [
    "@aws-lambda-powertools/logger",
    "@aws-lambda-powertools/tracer",
    "@aws-lambda-powertools/metrics",
  ],
});
```

_No response_

### PDK version used

0.11.3

_No response_

MacOS

cogwirrel:
Since in practice users are unlikely to want to name their service `example.hello#Hello`, we make `serviceName` a required field. This ensures that users define their own service name upfront, and the hello-world sample model will be generated with that name. Additionally, this change makes the documentation around changes to the `serviceName` clearer.

BREAKING CHANGE: `serviceName` is now a required property for Smithy projects. Additionally its components have been split into an object, for example `example.hello#HelloService` becomes `{ namespace: 'example.hello', serviceName: 'HelloService' }`

Fixes #176

### DIFF
--- a/packages/open-api-gateway/src/project/smithy/setup-smithy-build.ts
+++ b/packages/open-api-gateway/src/project/smithy/setup-smithy-build.ts
@@ -41,10 +41,8 @@ export const setupSmithyBuild = (
   options: SmithyApiGatewayProjectOptions
 ): SmithyBuildProjectResult => {
   const modelDir = options.modelDir ?? "model";
-  const fullyQualifiedServiceName =
-    options.serviceName ?? "example.hello#Hello";
-
-  const [serviceNamespace, serviceName] = fullyQualifiedServiceName.split("#");
+  const { namespace: serviceNamespace, serviceName } = options.serviceName;
+  const fullyQualifiedServiceName = `${serviceNamespace}#${serviceName}`;
 
   const smithyBuildDir = "smithy-build";
 
--- a/packages/open-api-gateway/src/project/smithy-api-gateway-java-project.ts
+++ b/packages/open-api-gateway/src/project/smithy-api-gateway-java-project.ts
@@ -49,7 +49,8 @@ export class SmithyApiGatewayJavaProject extends OpenApiGatewayJavaProject {
   ): OpenApiGatewayJavaProjectOptions {
     const { modelDir, generatedSpecFilePath } = setupSmithyBuild(
       this,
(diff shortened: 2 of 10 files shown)


packages/open-api-gateway/src/project/smithy/setup-smithy-build.ts
```diff
@@ -41,10 +41,8 @@ export const setupSmithyBuild = (
   options: SmithyApiGatewayProjectOptions
 ): SmithyBuildProjectResult => {
   const modelDir = options.modelDir ?? "model";
-  const fullyQualifiedServiceName =
-    options.serviceName ?? "example.hello#Hello";
-
-  const [serviceNamespace, serviceName] = fullyQualifiedServiceName.split("#");
+  const { namespace: serviceNamespace, serviceName } = options.serviceName;
+  const fullyQualifiedServiceName = `${serviceNamespace}#${serviceName}`;
```
agdimech, line 45: Shouldn't we validate these to ensure proper casing and usage of periods for namespaces?

cogwirrel: I actually started writing validation regexes but then figured it'd end up being validated by the Smithy compiler during smithy build anyway - so rather than repeat the validation a tad earlier I felt like it's simpler to just let Smithy do it :). The only downside with delegating I suppose is Smithy would error _after_ we've generated the model, so the user would need to either delete the model or update it.

agdimech (approved):

PR #189, opened 2022-10-07
fix(nx-monorepo): order yarn/npm/pnpm workspace packages in declaration order

cogwirrel:
Instead of alphabetical project name order followed by the additional workspace packages, order workspace packages by the order in which they are added to the monorepo. This allows for additional packages and subproject packages to be interlaced if necessary via the order of calls in the `.projenrc.ts`. This ensures the package manager can resolve local dependencies in the correct order as per the comment in issue #177

re #177


martenpayne (approved):
:+1: looks good

PR #193, opened 2022-10-29
fix(open-api-gateway): support primitive types in request and response body in lambda wrappers

cogwirrel:
Using primitive types, including `any`, for the top level type of the request/response body caused the TypeScript and Java lambda handler wrappers to fail to compile (and the Python one would fail at runtime).

Address this by ensuring no marshalling/demarshalling is performed when the request/response body is a primitive type, treating it as the raw string received in the api gateway event and returned in the api gateway result. This also ensures that the request/response body for primitive types is correctly typed as a string - previously eg an `application/pdf` request might be incorrectly typed as a `Blob` without it actually being coerced into one.

Responsibility is then delegated to the implementer of the lambda handler to transform the primitive type correctly.

### DIFF
--- a/packages/open-api-gateway/scripts/generators/python/templates/operationConfig.mustache
+++ b/packages/open-api-gateway/scripts/generators/python/templates/operationConfig.mustache
@@ -143,11 +143,11 @@ class {{operationIdCamelCase}}RequestArrayParameters(TypedDict):
 {{/allParams}}
     ...
 
-# Request body type (default to Any when no body parameters exist)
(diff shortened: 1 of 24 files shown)


packages/open-api-gateway/scripts/generators/python/templates/operationConfig.mustache
```diff
@@ -188,10 +201,24 @@ def {{operationId}}_handler(_handler: {{operationIdCamelCase}}HandlerFunction =
                 context,
                 interceptor_context,
             ), **kwargs)
+
+            response_body = ''
+            if response.body is None:
+                pass
```
agdimech, line 207: previously we would return '' as part of the response. Does passing achieve the same result?

agdimech (approved):

packages/open-api-gateway/scripts/generators/python/templates/operationConfig.mustache
cogwirrel: Yep, `response_body` will remain unchanged and so be used as `''`

PR #195, opened 2022-11-01
chore: workspace improvements

JeremyJonas:
Several workspace improvements for consistency and developer ergonomics
- chore(ide): add .vscode settings
- ci(pre-commit): only lint staged files during pre-commit hook
  > Prevent running extraneous linting against the entire workspace
and instead only lint the staged files that will be part of the
commit. 
  > Using a pre/post linting sha we can verify if any pre-commit processing
has modified the working tree and block the commit - as the user
should verify those changes first.
- fix(nx-monorepo): prevent extraneous subProject node installs
  > When synthesizing a new project for the first time, the package was not installed as a "yarn workspace" package, caused by the root package.json being cleaned up before the subprojects are generated. This caused several issues with dependencies between projects and within new projects given they were not initializing as a workspace package.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,6 +1,21 @@
 #!/bin/sh
+set -eo pipefail
+
 . "$(dirname "$0")/_/husky.sh"
 
+PRE_SHA="$(git diff HEAD --ignore-space-at-eol | shasum -a 256 --tag)"
+
 npx projen
-npx nx run-many --target=eslint --all && npx projen eslint
-git diff --ignore-space-at-eol --exit-code
+npx nx affected --target=eslint-staged --uncommitted
+npx projen eslint-staged
+
+POST_SHA="$(git diff HEAD --ignore-space-at-eol | shasum -a 256  --tag)"
+
+if [ "$PRE_SHA" != "$POST_SHA" ]
+then
+	RED='\033[0;31m'
+	NC='\033[0m' # No Color
+	echo "${RED}Pre-commit modified your working directory${NC} (eg: projen, eslint --fix)"
+	echo "${RED}Please verify and stage changed, then retry git commit${NC}"
+	exit 1
+fi
--- a/packages/nx-monorepo/src/nx-monorepo.ts
+++ b/packages/nx-monorepo/src/nx-monorepo.ts
@@ -25,7 +25,11 @@ import {
   TextFile,
   YamlFile,
 } from "projen";
-import { NodePackageManager, NodeProject } from "projen/lib/javascript";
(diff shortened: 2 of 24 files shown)


packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -242,7 +246,7 @@ export class NxMonorepoProject extends TypeScriptProject {
       release: options.release ?? false,
       jest: options.jest ?? false,
       defaultReleaseBranch: options.defaultReleaseBranch ?? "mainline",
-      sampleCode: false, // root should never have sample code
+      sampleCode: false, // root should never have sample code,
```
agdimech, line 249: why the trailing comma?

.husky/pre-commit
```diff
@@ -1,6 +1,21 @@
 #!/bin/sh
```
agdimech, line 1: nice :)

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -401,7 +405,31 @@ export class NxMonorepoProject extends TypeScriptProject {
...
+
+    // Prevent sub NodeProject packages from `postSynthesis` which will cause individual/extraneous installs.
+    // The workspace package install will handle all the sub NodeProject packages automatically.
+    const subProjectPackages: NodePackage[] = [];
+    this.subProjects.forEach((subProject) => {
+      if (isNodeProject(subProject)) {
+        const subNodeProject: NodeProject = subProject as NodeProject;
+        subProjectPackages.push(subNodeProject.package);
+        // @ts-ignore - `installDependencies` is private
+        subNodeProject.package.installDependencies = () => {};
```
agdimech, line 417: This is fantastic, although is there a way to achieve the same result using public API's?

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -401,7 +405,31 @@ export class NxMonorepoProject extends TypeScriptProject {
...
+      }
+    });
+
     super.synth();
+
+    // Force workspace install deps if any node subproject package has change, unless the workspace changed
+    if (
```
agdimech, line 424: same here, is there a way to achieve this using public APIs?

agdimech (changes requested):
This is a great PR, thanks for adding this Jeremy!

I noticed a build failure, I tried retriggering it to see if it was a transient failure.

agdimech:
Split out into https://github.com/aws/aws-prototyping-sdk/pull/196

PR #205, opened 2022-11-14
fix: change license header format to SPDX-License-Identifier

### LINKED ISSUE #199: fix: replace the verbose header text with shorthand SPDX-License-Identifier
## Legacy verbose header
https://github.com/aws/aws-prototyping-sdk/blob/1082c41443976f55a5c7fe8dff6c4e5e25a180a4/header.js#L1-L15

## Preferred SPDX header
```js
/*! Copyright [Amazon.com](http://amazon.com/), Inc. or its affiliates. All Rights Reserved.
SPDX-License-Identifier: Apache-2.0 */
```

## Additionally, use "block" rather than file for eslint header plugin options
https://github.com/aws/aws-prototyping-sdk/blob/1082c41443976f55a5c7fe8dff6c4e5e25a180a4/packages/identity/.eslintrc.json#L130-L132

"block" feature: https://github.com/Stuk/eslint-plugin-header#2-arguments

By using "block" we can prevent relative file issues with eslint, and since SPDX header is smaller it will not pollute the eslint config.

agdimech:
Change license header format to SPDX-License-Identifier.

Also adding an exclude for `samples/*` when running eslint-staged as this was causing nested sample packages to fail.

Fixes #199


JeremyJonas (approved):
Awesome, great work, thank you sir!

cogwirrel (approved):

PR #211, opened 2022-11-17
fix(docs): use roboto open-source font for docs

JeremyJonas:
Replace Amazon Ember usage with open-source Roboto

Need to confirm how build the docs and verify this - hence marking as Draft for now


JeremyJonas:
Verified Robot font is used in the doc build artifact and Ember is completely removed

image

agdimech (approved):

PR #216, opened 2022-11-21
fix: do not update snapshot or fix eslint when running in CI

agdimech:
Fix issue with snapshots being cached when they shouldn't be.

### DIFF
--- a/private/pdk-project.ts
+++ b/private/pdk-project.ts
@@ -13,7 +13,7 @@ import { Release } from "projen/lib/release";
 import {
   ProjectTargets,
   TargetDependencyProject,
-} from "../packages/nx-monorepo";
+} from "../packages/nx-monorepo/src";
 import {
   DEFAULT_NX_OUTPUTS,
   JEST_VERSION,
--- a/private/pdk-project.ts
+++ b/private/pdk-project.ts
@@ -10,6 +10,10 @@ import {
   Stability,
 } from "projen/lib/cdk";
 import { Release } from "projen/lib/release";
+import {
+  ProjectTargets,
+  TargetDependencyProject,
+} from "../packages/nx-monorepo";
 import { JEST_VERSION } from "./projects/pdk-monorepo-project";
 
 /**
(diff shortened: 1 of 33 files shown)


JeremyJonas (dismissed):

cogwirrel (dismissed):

private/pdk-project.ts
```diff
@@ -145,6 +149,33 @@ export class PDKProject extends JsiiProject {
...
     this.pdkRelease = new PDKRelease(this);
   }
+
+  /**
+   * Provides the ability to override project specific NX Project Targets.
+   *
+   * @return Nx ProjectTargets specific to this package.
+   */
+  public getNxProjectTargets(): ProjectTargets {
+    const relativeDir = `${this.outdir.split(this.root.outdir)[1]}`;
```
JeremyJonas: Is there reason not to use nx built-in var`{projectRoot}`? - https://nx.dev/more-concepts/customizing-inputs

private/projects/aws-prototyping-sdk-project.ts
```diff
@@ -146,4 +148,25 @@ export class AwsPrototypingSdkProject extends PDKProject {
 
     super.synth();
   }
+
+  /**
+   * @inheritDoc
+   */
+  getNxProjectTargets(): ProjectTargets {
+    const defaultTargets = super.getNxProjectTargets();
+    const relativeDir = `${this.outdir.split(this.root.outdir)[1]}`;
```
JeremyJonas: Should we expose this as a getter on PDKProject? Seems useful

JeremyJonas (dismissed):
Stellar fix and improvements, thank you!

private/pdk-project.ts
agdimech: DIdn't know this existed! Will refactor :)

private/projects/aws-prototyping-sdk-project.ts
agdimech: Ill keep as is for now as it's an internal construct. If needed we can make public later.

JeremyJonas (dismissed):

JeremyJonas (approved):

JeremyJonas (approved):

PR #224, opened 2022-11-29
fix(nagpack): Corrected errors in initial AwsPrototyping NagPack rule scoping

dboyd13:
Fixed issue where initial commit of `AwsPrototyping` NagPack included rules beyond the intended scope of pack

Impacts `WARN` scope only.


agdimech (commented):
Why are these rules being removed?

dboyd13:
> Why are these rules being removed?

Do not meet the intended Pack scope. Brought into initial commit, due to human error.

agdimech (approved):

PR #231, opened 2022-12-02
chore: improve nx config support and caching

JeremyJonas:
### Improve NX project and target configurations
- enable PDKProject direct nx project config
- move default nx build target to workspace
- remove legacy nx override logic
- separate out aws-arch generate task and enable caching

### Improve aws-arch generator caching
- Replace "cache dir" with "hash based" temp dir to ensure unique empty dir while somewhat minimizing requests during dev
- Add `static/**` to generate input caching


packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -458,10 +345,7 @@ export class NxMonorepoProject extends TypeScriptProject {
       .forEach((subProject: Project) => {
         // generate a package.json if not found
         const manifest: any = {};
-        (manifest.nx = this.targetOverrides[subProject.outdir]
-          ? { targets: this.targetOverrides[subProject.outdir] }
-          : undefined),
-          (manifest.name = subProject.name);
+        manifest.name = subProject.name;
```
agdimech: why is this needed?

packages/nx-monorepo/src/nx-types.ts
```diff
@@ -0,0 +1,186 @@
...
+  export interface TargetDefaults {
+    readonly [name: string]: ProjectTarget;
+  }
+
+  /**
+   * NX workspace configurations.
+   *
+   * @see https://nx.dev/configuration/packagejson
+   */
+  export interface WorkspaceConfig {
```
agdimech, line 98: Is this a breaking change given previously this was called NXConfig?

agdimech (commented):
Overall it looks good and is much cleaner. Just had a few comments.

packages/nx-monorepo/src/nx-types.ts
JeremyJonas: Yes it is breaking, but only low-level config within nx-monorepo which should only be used by internals. As discussed I have updated the nx-monorepo README to include breaking details, along with commit message to denote such as well.

Given this is still v0, it is expected to break and we have now added sufficient messaging for potential external users.

JeremyJonas:
@agdimech I have re-opened this stale/closed PR, which resolves build issues for https://github.com/aws/aws-prototyping-sdk/pull/266. 

I have rebased mainline and updated the commits as discussed - ready for your review.

JeremyJonas:
@agdimech After digging a bit deeper here, the real issue is the `aws-arch` generate script is simply not deterministic due to not being able to pin pricing manifest content that is fetch from the url. All the other sources are hashed urls, so deterministic between builds.

The only way I could make this deterministic is to commit the pricing-manifest.json file to the source code.

I also added a github workflow to run weekly to auto fetch the pricing and run build (which also updates the snapshots) - currently it requires manual approval since I want to see the snapshots since there is lots of auto mapping atm.

Hopefully this works well enough.

Commit for this new code here: https://github.com/aws/aws-prototyping-sdk/pull/231/commits/26112111c16b07e61a25991cd490cf7a8fb0fdc4

agdimech (approved):

agdimech (approved):

Issue #238, opened 2022-12-07
[FEATURE] open-api-gateway - allow all operations to use the same lambda integration 


mteichtahl:
### Describe the feature

Today an ```Api``` requires a lambda function and associated integration for each operation. This proposed feature should allow a developer to send all operations to a single lambda integration.

### Use Case

As a developer 
- [ ] I want to use a single lambda as a handler for all ```Api``` operations.
- [ ] PDK should provision my single lambda 
- [ ] the provisioned lambda should provide me a simple way of integrating my modules for each operation
- [ ] I want this functionality to work regardless of my lambda implementation language

### Proposed Solution

```typescript
export class SampleApi extends Api {
  constructor(scope: Construct, id: string) {
    super(scope, id, {
      defaultAuthorizer: Authorizers.iam(),
      integration:  Integrations.lambda(new NodejsFunction(scope, 'apiHandlerFunction')
    });
  }
}```

### Other Information

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### PDK version used

0.12.24

### What languages will this feature affect?

Typescript, Java, Python

### Environment details (OS name and version, etc.)

N/A

**Comment by cogwirrel** [comment]
I've raised a PR for this with only TypeScript support for now. Rather than opt for a single `integration` property, I kept the existing construct interface the same and provided a helper method `Operations.all` which can repeat the integration for every operation, eg:

```ts
import { Operations } from 'my-api-typescript-client';

export class SampleApi extends Api {
  constructor(scope: Construct, id: string) {
    super(scope, id, {
      defaultAuthorizer: Authorizers.iam(),
      integrations: Operations.all({
        integration: Integrations.lambda(new NodejsFunction(scope, 'apiHandlerFunction')),
      })
    });
  }
}
```

In your lambda function, you'd export the `handlerRouter` as your handler function, which can delegate to your type-safe handlers for each operation, eg:

```ts
import { handlerRouter, sayHelloHandler, sayGoodbyeHandler } from "my-api-typescript-client";
import { corsInterceptor } from "./interceptors";
import { sayGoodbye } from "./handlers/say-goodbye";

const sayHello = sayHelloHandler(async ({ input }) => {
  return {
    statusCode: 200,
    body: {
      message: `Hello ${input.requestParameters.name}!`,
    },
  };
});

export const handler = handlerRouter({
  // Interceptors declared in this list will apply to all operations
[...]

PR #240, opened 2022-12-08
feat(open-api-gateway): generate a router for using a single lambda function for all operations

cogwirrel:
The generated API client now includes a `handlerRouter` which can be used to route requests to the appropriate handler if users desire to deploy a single shared lambda for all API integrations.
    
In TypeScript, this is type-safe, ensuring all operations are mapped to a handler produced by the appropriate handler wrapper method.
    
Additionally we supply a convenience method `Operations.all` to make it easy to share the same integration for all operations.
    
This change includes support for TypeScript and Python - Java coming soon!

re #238

### DIFF
--- a/packages/open-api-gateway/README.md
+++ b/packages/open-api-gateway/README.md
@@ -529,7 +529,6 @@ required:
 A sample construct is generated which provides a type-safe interface for creating an API Gateway API based on your OpenAPI specification. You'll get a type error if you forget to define an integration for an operation defined in your api.
 
 ```ts
-import * as path from 'path';
 import { Authorizers, Integrations } from '@aws-prototyping-sdk/open-api-gateway';
 import { NodejsFunction } from 'aws-cdk-lib/aws-lambda-nodejs';
 import { Construct } from 'constructs';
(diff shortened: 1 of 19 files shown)


packages/open-api-gateway/README.md
```diff
@@ -553,6 +552,30 @@ export class SampleApi extends Api {
...
+
+export class SampleApi extends Api {
+  constructor(scope: Construct, id: string) {
+    super(scope, id, {
+      defaultAuthorizer: Authorizers.iam(),
+      // Use the same integration for every operation.
+      integrations: Operations.all({
+        integration: Integrations.lambda(new NodejsFunction(scope, 'say-hello')),
+      }),
+    });
```
JeremyJonas, line 574: Might be better to support a "default integration", which would allow to specify a uniq integration for specific routes while using the "default" (Operations.all) for the rest.

With current implementation, say I have 100 routes of which "99 share integration" and "1 has uniq integration", do I have to list all 100 because 1 of them does not use the shared?

JeremyJonas (approved):
Looks great Jack. Would be nice to address the comment around "default vs all", but that can be added later.

Also would be really great to replace mustache Typescript AST and JSII (auto-generate python+java++), but likely needs to wait for JSII to [support Typescript 4.x](https://github.com/aws/jsii/issues/3609) to handle the interfaces. I will create a feature request for that to track.

packages/open-api-gateway/README.md
cogwirrel: I thought about this a little but didn't want to remove the type-safety in TypeScript where it forces you to provide an integration for every operation 😄.

I figured you can still achieve this use case fairly easily with something like:

```ts
integrations: {
  ...Operations.all({ integration: myIntegration }),
  otherOperation: { integration: myOtherIntegration },
},
```

In python it's a little uglier (but I fully blame python for that 😉 ):

```python
integrations = Operations.all(
  OpenApiIntegration(
    integration=my_integration
  ),
)
integrations.other_operation = OpenApiIntegration(
  integration=my_other_integration
)

self.api = Api(self, 'Api',
  default_authorizer=Authorizers.iam(),
  integrations=integrations,
)
```

cogwirrel:
> Looks great Jack. Would be nice to address the comment around "default vs all", but that can be added later.
> 
> Also would be really great to replace mustache Typescript AST and JSII (auto-generate python+java++), but likely needs to wait for JSII to [support Typescript 4.x](https://github.com/aws/jsii/issues/3609) to handle the interfaces. I will create a feature request for that to track.

Yes this would be so good!! It does get a little tedious having to implement every change to generated code in all three languages! 😆

packages/open-api-gateway/README.md
JeremyJonas: That works pretty well actually, much cleaner than I expected. Mostly just wanted to know it was possible, and with your example it definitely is - great work!

JeremyJonas: I created discussion https://github.com/aws/aws-prototyping-sdk/discussions/241 for the mustache to AST/JSII idea.

PR #242, opened 2022-12-09
feat(open-api-gateway): java support for router for using single lambda function for all operations

### LINKED ISSUE #238: [FEATURE] open-api-gateway - allow all operations to use the same lambda integration

cogwirrel:
Adds a `HandlerRouter` class to the generated Java client which can be extended and used as the lambda entrypoint for all operations.

The router requires you to define a method for every operation, which returns a handler for that operation. Like with TypeScript and Python, these handlers are the same handlers can be used in a standalone fashion, so the router is another layer that can be easily added/removed should users decide to consolidate/split their lambdas.

The router can also be decorated with interceptors to apply to all operations (and/or override the `getInterceptors` method to return interceptors which need parameters for construction).

fix #238

### DIFF
--- a/packages/open-api-gateway/README.md
+++ b/packages/open-api-gateway/README.md
@@ -834,6 +834,35 @@ public class SayHelloHandler extends SayHello {
 }
 ```
 
+##### Handler Router
+
+The lambda handler wrappers can be used in isolation as handler methods for separate lambdas. If you would like to use a single lambda function to serve all requests, you can do so by extending the `HandlerRouter` class.
+
+```java
+import com.generated.api.myapijava.client.api.Handlers.SayGoodbye;
+import com.generated.api.myapijava.client.api.Handlers.HandlerRouter;
+import com.generated.api.myapijava.client.api.Handlers.Interceptors;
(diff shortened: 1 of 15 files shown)


JeremyJonas (approved):
LGTM - Nice work Jack! Very clean

PR #244, opened 2022-12-09
fix: bundle hoisting and cdk-graph packaging

JeremyJonas:
`bundledDependencies` where not published for `cdk-graph*` packages due to them getting hoisted by workspace. This PR fixes this at the workspace level by applying "nohoist" automatically for all sub-project bundled deps. And patches the postinstall issues for `cdk-graph-plugin-diagram` package.

- fix(nx-monorepo): do not hoist bundled deps of sub-projects
- feat(pdk-project): support disabling pypi+maven publishing
    > For experimental packages it is best to ensure Typescript version is working sufficiently before publishing to other languages. This feat just adds ability to mark as `false` to prevent publishing via code.
- fix(cdk-graph): only publish cdk-graph-x to npm
    > Until Typescript version is more stable and tested I am preventing pulishing of `cdk-graph` and `cdk-graph-plugin-diagram` to pypi and maven. Once have fixed bundling issues and verify with other users that the
Typescript version is working I will re-enable this.
- fix(cdk-graph): resolve cdk-graph+plugin-diagram deps and postinstall
    > postinstall is npm lifecycle so can't use projen tasks as they are not available


private/pdk-project.ts
```diff
@@ -28,14 +28,14 @@ export interface PDKProjectOptions extends JsiiProjectOptions {
    *
    * @default - package will be published with module name: aws_prototyping_sdk.<your_package_name>
    */
-  readonly publishToPypiConfig?: JsiiPythonTarget;
+  readonly publishToPypiConfig?: JsiiPythonTarget | false;
```
cogwirrel, line 31: The default behaviour from before of publishing when `undefined` is bit weird, and it feels a bit strange having different behaviour for different falsy values :) - I wonder if we should change this interface to something a bit more explicit?

```ts
interface PublishConfig<T> {
  readonly enabled: boolean;
  readonly T?: options;
}
```

It's a breaking change but only internally in PDK I think :)

public/docs/scripts/generate-markdown.js
```diff
@@ -33,6 +33,19 @@ function includeBanner(pkg, markdown, stability) {
   return stability !== 'stable' ? `${generateExperimentalBanner(pkg)}\n${markdown}`: markdown;
 }
 
+function isPgkLanguageTarget(language, jsii) {
```
cogwirrel: Nit: pkg? :)

private/projects/pdk-monorepo-project.ts
```diff
@@ -95,14 +95,6 @@ export class PDKMonorepoProject extends NxMonorepoProject {
...
           "@aws-prototyping-sdk/*/oss-attribution-generator",
           "@aws-prototyping-sdk/*/oss-attribution-generator/*",
-          "@aws-prototyping-sdk/open-api-gateway/openapi-types",
-          "@aws-prototyping-sdk/open-api-gateway/openapi-types/*",
-          "@aws-prototyping-sdk/open-api-gateway/fs-extra",
-          "@aws-prototyping-sdk/open-api-gateway/fs-extra/*",
-          "@aws-prototyping-sdk/open-api-gateway/lodash",
-          "@aws-prototyping-sdk/open-api-gateway/lodash/*",
-          "@aws-prototyping-sdk/open-api-gateway/log4js",
-          "@aws-prototyping-sdk/open-api-gateway/log4js/*",
```
cogwirrel, line 105: Looks like the bundled deps for `open-api-gateway` aren't being picked up?

cogwirrel (commented):
Looks great! Will be nice not to have to remember the nohoist! :D A couple of small comments :)

private/projects/pdk-monorepo-project.ts
JeremyJonas: Oh weird, they were added, must have changed something with cleanup refactor. Ok will fix this, good catch!

private/pdk-project.ts
JeremyJonas: Agree, and these PDK publish option proxies also disable the default projen publish options... which I assume is to just enable defaults by default. 

I created https://github.com/aws/aws-prototyping-sdk/discussions/246 to discuss similar concerns. Based on decision from there, I think we should refactor how we handle this. 

Maybe just remove the current PDK publish options, use the default projen options, but add new explicit "publishX" Boolean options (with default based on poll results). 

For now I think how the PR does it is ok, as just provides way to disable, which currently is available. Then we can fix properly is follow up.

cogwirrel: Cool that sounds like a plan, happy to leave as is for now and revisit with that issue :)

private/projects/pdk-monorepo-project.ts
JeremyJonas: @cogwirrel Actually this is correct. They are removed from `pdk-monorepo-project`, but still in `package.json`. Because of the "automatic nohoist for bundles" they are still in package.json, and no longer need to be maintained in the projen project :) 

https://github.com/aws/aws-prototyping-sdk/pull/244/files/f26185dc614c6194b9a6933ea2ded257a8a99467#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

image

> The cloudscape stuff is also not removed as per the screenshot, just moved higher in the list - see actual package.json file

cogwirrel: ahh yes of course! thank you! :D

cogwirrel (approved):

PR #253, opened 2022-12-15
feat(pipeline): compile ts pipeline as part of build by default

cogwirrel:
The `AwsCdkTypeScriptApp` which the `PDKPipelineTsProject` inherits from disables compilation due to synth executing via TypeScript anyway. (See https://github.com/projen/projen/issues/1958)

However, any `NodejsFunction`s that are referenced won't be type checked at build time which can lead to mistakes. We err on the side of caution by adding TypeScript compilation to the build by default, given `NodejsFunction` is used quite frequently.


martenpayne (approved):
Looks good 👍

PR #255, opened 2022-12-16
chore: upgrade minimatch/loader-utils for dependabot alerts (really)

cogwirrel:
Resolutions only take effect in the top level package.json of the monorepo, so adjust to ensure dependabot alerts are resolved.

NB the resolutions can likely be removed with an upgrade of `react-scripts` once the following is addressed: https://github.com/facebook/create-react-app/issues/12851


martenpayne (approved):
all good 👍

Issue #269, opened 2023-01-20
[FEATURE] Smithy should provide a way to group API operators in generated documentation


mteichtahl:
### Describe the feature

Today, when documentation is generated from smithy it doesn't respect/understand the needs of documentation generation tools. For example, redoc, requires an x-tag for operator grouping - see https://redocly.com/docs/api-reference-docs/specification-extensions/x-tags/

### Use Case

support a mechanism via custom tags or other implementation so that documentation generated is presentable and usable to customers. 

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### PDK version used

0.12.24

### What languages will this feature affect?

Typescript, Java, Python

### Environment details (OS name and version, etc.)

all

cogwirrel:
Had a quick look at this - unfortunately the `x-tags` property only seems to apply to schemas in redoc, not operations :(. Otherwise it would be possible to add these tags using the [`jsonAdd` smithy build option](https://smithy.io/2.0/guides/converting-to-openapi.html#converting-to-openapi-with-smithy-build), eg:

```typescript
new SmithyApiGatewayTsProject({
  ...
  smithyBuildOptions: {
    projections: {
      openapi: {
        plugins: {
          openapi: {
            jsonAdd: {
              '/paths/~1hello/get/x-tags': ['foo', 'bar'],
            },
          },
        },
      },
    },
  },
});
```

It seems like redoc (and other openapi documentation generators) usually group operations by their `tags` property. Smithy can preserve tags from the `@tags` trait when generating openapi with the `tags` [smithy build option](https://smithy.io/2.0/guides/converting-to-openapi.html#converting-to-openapi-with-smithy-build). But this causes other issues unfortunately.

If you introduce tags, the code generators treat each tag as a separate "API", meaning a different client class is generated per-tag. This is problematic as it means the custom lambda handlers and operation lookup etc generated by the open-api-gateway package are also generated per-api, and at present the package assumes there is only one. It might be possible to update the codegen templates to address this but it needs a bit more investigation!

cogwirrel:
As an update here, updating the lambda handler wrapper generators to support tags is non-trivial since operations may be duplicated across multiple APIs (ie. can be tagged with more than one tag) - so we require some kind of de-duping of operations to make sure it works for all cases.

[...]

PR #276, opened 2023-02-02
feat(open-api-gateway): support singular tags on operations

cogwirrel:
This change adds support for tagging operations, for example:

```yaml
  /hello:
    get:
      tags:
        - foo
      operationId: sayHello
      responses:
        '200':
          description: Ok
```

Previously, tagging would cause broken lambda handler wrappers etc to be generated since the CDK construct assumes a single `OperationConfig`, and tagging APIs caused multiple to be generated.

In this change, we always generate a single `OperationConfig`, but iterate over all apis to make sure all operations for all tags are included.

⚠️ Note that this change only supports tagging each operation with __zero or one__ tags, multiple tags will cause duplicated code to be generated which will not build - support for multiple tags requires an upstream change in openapi-generator: https://github.com/OpenAPITools/openapi-generator/pull/14568

We change the default configuration for the Smithy openapi projection to preserve tags since if they are specified in the model users would expect them to apply to the generated code/documentation etc. This means that operations can be tagged in the Smithy model like so:

```smithy
@readonly
@http(method: "GET", uri: "/hello")
@tags(["foo"])
operation SayHello {
    input: SayHelloInput
    output: SayHelloOutput
    errors: [ApiError]
}
```

This behaviour can be overridden in the project's `smithyBuildOptions`, ie:

```ts
const api = new SmithyApiGatewayTsProject({
  ...
  smithyBuildOptions: {
    projections: {
      openapi: {
        plugins: {
          openapi: {
            tags: false,
          },
        },
      },
    },
  },
});
```

re #269


packages/open-api-gateway/test/project/codegen/components/generated-java-client-source-code.test.ts
```diff
@@ -28,4 +28,8 @@ describe("Generated Java Client Code Unit Tests", () => {
   it("Multi", () => {
     expect(synthGeneratedCode("multi.yaml")).toMatchSnapshot();
   });
+
+  it("Single Tags", () => {
+    expect(synthGeneratedCode("single-tags.yaml")).toMatchSnapshot();
+  });
 });
```
JeremyJonas: What does the error look like if use inputs "multiple tags"? Are we able to catch this early in the build and output detailed error indicating multiple is not supported at this time?

Can we add a unit test for multiple tags as well, which is expected to throw an error with actionable messaging?

cogwirrel: Good call! The error currently would be after all the code is generated (`npx projen`), and would surface as a build error for the generated client (eg defining two methods with the same name due to operations being repeated across different APIs). To test that case properly we'd need to bring back the verdaccio inegration tests I think.

I think we could catch this at `npx projen` time by adding some validation on the openapi spec (for smithy this would be after the openapi spec is generated) - I'll have a look and see what I can do :)

cogwirrel: Added some validation :)

JeremyJonas (approved):
Looks awesome, great work adding the validation and actionable error handling :D

PR #280, opened 2023-02-03
feat(open-api-gateway): idomatic smithy project structure for better IDE support

### LINKED ISSUE #278: [FEATURE] Support Smithy IntelliJ plugin
### Describe the feature

We should support the [Smithy IntelliJ Plugin](https://github.com/iancaffey/smithy-intellij-plugin) to improve the user experience when working with Smithy.

### Use Case

The best I could set up in intellij was syntax highlighting by [using the vs-code plugin as a textmate bundle](https://github.com/awslabs/smithy-vscode#use-with-intellij), but this doesn't have proper auto-complete and error checking, making the developer experience worse than it could be.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

- [X] This feature might incur a breaking change

### PDK version used

0.13.4

Typescript, Java, Python

OSX

cogwirrel:
This change restructures the smithy-based projects to combine the `model` and `smithy-build` directories into a single `smithy` directory. This directory has a more similar structure to those found in the Smithy documentation, and in turn means that IDE tools can be more easily configured to help with the developer experience.

Specifically with this change, the Smithy IntelliJ plugin can be used to enhance the development experience when authoring models. https://github.com/iancaffey/smithy-intellij-plugin

We preserve the behaviour of abstracting away the gradle wrapper, whereby it is ignored from the generated project's source control by default, and copied from the PDK if it does not already exist.  Similarly we allow users to bring their own `gradle wrapper` by choosing not to ignore it in source control, and placing it in the `smithy` directory. This replaces the `gradleWrapperPath` configuration option.

While this is a breaking change, the migration steps are fairly straightforward:

- Upgrade the open-api-gateway package
- `npx projen` to regenerate code - notice the new `smithy` folder in your api project
- Copy the contents of the old `model` directory into `smithy/src/main/smithy`
- If any custom dependencies were added to the `smithy-build/build.gradle` file, add them to `smithy/build.gradle`
- Delete the old `model` and `smithy-build` directories

BREAKING CHANGE: Moved the location of the smithy model from 'model' to 'smithy/src/main/smithy'. Removed the gradleWrapperPath configuration option.

fix #278


packages/open-api-gateway/src/project/smithy/smithy-build-project.ts
```diff
@@ -3,21 +3,19 @@ SPDX-License-Identifier: Apache-2.0 */
...
   /**
-   * Fully qualified service name
+   * Smithy service name
    */
-  readonly fullyQualifiedServiceName: string;
-  /**
-   * Absolute path to the model directory
-   */
-  readonly modelPath: string;
+  readonly serviceName: SmithyServiceName;
```
JeremyJonas, line 18: This is a breaking change as well right? Given that these options are public interface for the project?

packages/open-api-gateway/README.md
```diff
@@ -100,9 +100,10 @@ new SmithyApiGatewayTsProject({
 This will result in a directory structure similar to the following:
 
 ```
-|_ model/
-    |_ main.smithy - The Smithy model used to define your API. You can define as many .smithy files in here as you like.
-|_ smithy-build/
+|_ smithy/
+    |_ src/
+        |_ main/
+            |_ smithy - The Smithy model used to define your API. You can define as many .smithy files in here as you like.
```
JeremyJonas, line 106: We should log the breaking changes and how to migration in the README.

Example:
https://github.com/aws/aws-prototyping-sdk/blob/mainline/packages/nx-monorepo/README.md?plain=1#L5-L6
```markdown
> **BREAKING CHANGES** (pre-release)
> - v0.13.0: `WorkspaceConfig.nxConfig` type `NxConfig => Nx.WorkspaceConfig`, and `overrideProjectTargets` removed in favor of `PDKProject.nx` config to manage Nx project configurations. See [PR 231](https://github.com/aws/aws-prototyping-sdk/pull/231).
```

JeremyJonas (changes requested):
Is it possible to add auto-migration capabilities - thinking for the model folder move would be great. More curious from how we can handle migrations in general for packages that this particularly since quite minimal. Nothing have to do now, but could be a nice simple example to attempt 😄

packages/open-api-gateway/src/project/smithy/smithy-build-project.ts
cogwirrel: This project is internal so I think this is ok! Though can technically get at it via `import { SmithyBuildProject } from "@aws-prototyping-sdk/open-api-gateway/lib/project/smithy/smithy-build-project";`...

packages/open-api-gateway/README.md
cogwirrel: Good call, will update!

cogwirrel:
> Is it possible to add auto-migration capabilities - thinking for the model folder move would be great. More curious from how we can handle migrations in general for packages that this particularly since quite minimal. Nothing have to do now, but could be a nice simple example to attempt 😄

This would be super cool, but is a little bit fiddly to do properly as we'd need to parse the old `build.gradle` file and copy the dependencies across into the new one. Given we're still at 0.x I think it's more trouble than it's worth! 😄

packages/open-api-gateway/src/project/smithy/components/sample-executable.ts
```diff
@@ -0,0 +1,38 @@
...
+import * as fs from "fs";
+import * as path from "path";
+import { Project, SampleFile, SampleFileOptions } from "projen";
+import { getFilePermissions } from "projen/lib/util";
+
+/**
+ * A sample file that is executable
+ */
+export class SampleExecutable extends SampleFile {
+  private readonly file: string;
```
agdimech: minor: rename to filePath

packages/open-api-gateway/src/project/smithy/components/sample-executable.ts
```diff
@@ -0,0 +1,38 @@
...
+  constructor(project: Project, filePath: string, options: SampleFileOptions) {
+    super(project, filePath, options);
+    this.file = filePath;
+  }
+
+  /**
+   * @inheritDoc
+   */
+  public synthesize() {
+    const fullFilePath = path.join(this.project.outdir, this.file);
```
agdimech: Should we just store this in the private filePath variable as part of the constructor?

packages/open-api-gateway/src/project/smithy/components/sample-executable.ts
cogwirrel: I named it that first but jsii complained because I was overriding a subclass property haha :)

packages/open-api-gateway/src/project/smithy/components/sample-executable.ts
cogwirrel: Can do! :)

cogwirrel: Will name it `fullFilePath` and kill 2 birds with one stone :)

agdimech (approved):

JeremyJonas (approved):

Issue #170, opened 2022-09-26
[FEATURE] Support smithy-vscode extension


sperka:
### Describe the feature

To create a projen-managed artifact that is generated when one uses a `smithy-api-gateway-XXX-project` in order to support smithy vscode extension and not to fail with validations that could be easily avoided.

### Use Case

In the current form, `smithy-vscode` extension is constantly failing with the error of not recognizing `aws.protocols#restJson1` trait:

```
Caused by: software.amazon.smithy.model.validation.ValidatedResultException: Result contained ERROR severity validation events: 
[ERROR] my.namespace#MyService: Unable to resolve trait `aws.protocols#restJson1`. If this is a custom trait, then it must be defined before it can be used in a model. | Model ....
```

This becomes annoying after some time.

### Proposed Solution

As the [README](https://github.com/awslabs/smithy-vscode/blob/main/README.md#authoring-a-model) says, a `smithy-build.json` should be present in the root of the project.

Generating a `smithy-build.json` with the following content in the root of the project solves the issue mentioned above:
```json
{
  "version": "2.0",
  "maven": {
    "dependencies": [
      "software.amazon.smithy:smithy-cli:1.24.0",
      "software.amazon.smithy:smithy-model:1.24.0",
      "software.amazon.smithy:smithy-openapi:1.24.0",
      "software.amazon.smithy:smithy-aws-traits:1.24.0"
    ],
    "repositories": [{ "url": "https://repo1.maven.org/maven2/" }]
  },
  "//": "~~ Generated by projen. To modify, edit .projenrc.js and run \"npx projen\"."
}
```

**NOTE**: the dependencies can be generated based on the smithy project's settings in the PDK project. More specifically, all the dependencies that are present in `packages/project-name/smithy-build/build.gradle`'s `dependencies` section, can be added to the "root" `smithy-build.json`. This won't have any effect generating the API project.

Additionally, based on the smithy-vscode codebase, there is a way of defining a `rootPath` for the extension: [source](https://github.com/awslabs/smithy-vscode/blob/main/src/extension.ts#L136). It would be ideal to generate a `.smithy/smithy-build.json` file, but unfortunately at this point `smithyLsp.rootPath` in `.vscode/settings.json` results in `Unknown Configuration Setting` and the value is not picked up by the Smithy language server.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### PDK version used

0.11.2

### What languages will this feature affect?

Typescript

[...]

PR #285, opened 2023-02-09
feat(open-api-gateway): include maven dependencies in smithy-build.json

cogwirrel:
Move `build.gradle` to a managed generated file, and render maven dependencies into `build.gradle` and `smithy-build.json`. This should mean we're a step closer to the vs code plugin working, since unlike the intellij plugin it looks at `smithy-build.json` rather than `build.gradle` to resolve its dependencies.

Upgrades projen to `^0.67.41` which includes https://github.com/projen/projen/pull/2412

re #170


cogwirrel:
As per https://github.com/aws/aws-prototyping-sdk/issues/170 the vs-code plugin still requires `smithy-build.json` to be at the project root, but with this change the plugin has everything it needs to work properly - and works as expected when you open just the `smithy` folder :)

eg. hovering over the imported `restJson1` trait:

vscode screenshot

It also seems to work ok if you copy/paste `smithy-build.json` to the root of a monorepo - so I assume symlinking as part of project synthesis would also work - this would need to be a customer step though, as we don't want to create files outside of the bounds of the project, and customers would need to pick one if they have multiple `SmithyApiGatewayXXXProject`s in their monorepo.

agdimech (approved):

PR #286, opened 2023-02-10
feat: add support for YARN2 packagemanager in open-api-gateway

sperka:
Fixes #

* use projen version `^0.67.44`, with the new  option `NodePackageManager.YARN2`
* support `YARN2` in `open-api-gateway`
* fix issue in `open-api-gateway` introduced by `volta` preemptively adding symlink to non-existent `pnpm`
* `open-api-gateway` generates specs with package manager preference `pnpm` > `yarn` > `npm`


cogwirrel (approved):
Nice work, thank you! 🚀

PR #297, opened 2023-02-22
fix(open-api-gateway): fix service name docs and remove unused modelDir option

cogwirrel:
Not counting this as a breaking change as it was already broken in https://github.com/aws/aws-prototyping-sdk/pull/280 (which was marked as breaking 😄 ).


agdimech (commented):
There are additional places in the README where serviceName is not present. Please add to all other locations.

agdimech:
On a seperate note: Can we disable jest for the GeneratedTypescriptClientProject given no tests are present?

agdimech (approved):

PR #306, opened 2023-03-09
fix(cdk-graph-plugin-diagram): resolve sharp rebuild issues

JeremyJonas:
Fixes #


private/projects/cdk-graph-plugin-diagram.ts
```diff
@@ -68,11 +66,28 @@ export class CdkGraphPluginDiagramProject extends CdkGraphPluginProject {
...
+            "npx -p @aws-prototyping-sdk/nx-monorepo nx-dir-hasher {workspaceRoot}/packages/cdk-graph-plugin-diagram/node_modules/sharp",
+        },
+      ],
+      true
+    );
+    this.nxOverride(
+      "targets.build.outputs",
+      ["{projectRoot}/node_modules/sharp"],
+      true
+    );
```
JeremyJonas, line 90: This is the core fix to the "sharp prebuild" issue, the rest is to properly support this.
- To update default "inputs" we needed to define "default" namedInput
- Nx wil **always** ignore all node_modules/* path, so had the create custom `nx-dir-hasher` supports hashing paths within node_modules
- Needed to remove CI conditional for prebuild so that local development can benefit from remote caching

private/projects/cdk-graph-plugin-diagram.ts
```diff
@@ -57,8 +57,6 @@ export class CdkGraphPluginDiagramProject extends CdkGraphPluginProject {
       stability: Stability.EXPERIMENTAL,
     });
 
-    this.addPackageIgnore("**/node_modules");
```
JeremyJonas, line 60: This has no effect, as already managed by `npm pack`

packages/aws-arch/scripts/generate-assets.ts
```diff
@@ -19,7 +19,7 @@ import * as util from 'util';
 import * as stream from 'stream';
 import * as fs from 'fs-extra';
 import fetch from 'node-fetch';
-import { Parse as ZipParse } from 'unzipper';
+import { Parse as ZipParse,  } from 'unzipper';
```
agdimech: redundant comma

packages/cdk-graph-plugin-diagram/package.json
```diff
@@ -198,6 +198,36 @@
...
+          {
+            "target": "build",
+            "projects": "dependencies"
+          }
+        ],
+        "inputs": [
+          "default",
+          "^default",
+          {
+            "runtime": "npx -p @aws-prototyping-sdk/nx-monorepo nx-dir-hasher {workspaceRoot}/packages/cdk-graph-plugin-diagram/node_modules/sharp"
```
agdimech, line 225: I assume this invalidates the caches if the nested sharp dep changes?

JeremyJonas: Yes that is correct. It includes the hash of the `node_modules/sharp` directory in the caching hash, which if changed will re-build.

Along with the follow lines that will ensure sharp build files are output from cache as well
```ts
    this.nxOverride(
      "targets.build.outputs",
      ["{projectRoot}/node_modules/sharp"],
      true
    );
```

packages/aws-arch/scripts/generate-assets.ts
JeremyJonas: Fixed

agdimech (approved):

Issue #282, opened 2023-02-06
[BUG] dependencies are not installed by npm


tmokmss:
### Describe the bug

Hi, I'm using PDK with npm and observing a weird behavior about npm install or npm update. Could you check this? The reproduction steps are below.

### Expected Behavior

`npm i aws-prototyping-sdk` should install all the dependencies, and `npm update` won't remove any packages.

### Current Behavior

Not all the required packages are installed on `npm i aws-prototyping-sdk`.

Also `npm update` removes several dependent packages of PDK, causing inconsistency between package.json and package-lock.json. 

### Reproduction Steps

Run the following commands in an empty directory:

```sh
npm init -y
npm i aws-prototyping-sdk
npm ci
# -> npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.

npm i
npm ci
# -> success

npm update
npm ci
# -> npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

I confirmed the issue on both npm v8 and v9.

### Additional Information/Context

`npm ci`  is the command we recommend for customers to use to install dependencies, but currently customers sometimes see the above error, getting confused. (e.g. after running `npm update`, they get the error on `npm ci`.)

### PDK version used

0.13.5

### What languages are you seeing this issue on?

TypeScript

### Environment details (OS name and version, etc.)

macOS 12.6.3

agdimech:
This has been resolved in #315

tmokmss:
Hi @agdimech thank you for working on the issue! I'm afraid it is still not fixed though. I see the same npm error by the same reproduction step:

```
npm init -y
npm i aws-prototyping-sdk
npm ci
# expecting npm ci successes, but we get the below error:
# -> npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

Reproducible with `aws-prototyping-sdk: 0.14.16`.

tmokmss:
Hi @agdimech I did a bit of research about this. The below is the command history when reproducing this issue:

```
$ npm install aws-prototyping-sdk

added 8 packages, and audited 119 packages in 20s

11 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

$ npm install

[...]

PR #315, opened 2023-03-21
refactor: support pnpm

### LINKED ISSUE #308: [BUG] nx cache location change not synced with gitignore
### Describe the bug

In previous versions default `nx` cache directory was at `node_modules/.cache/nx`, which by default was gitignored.
In a current [change](https://github.com/nrwl/nx/blob/46b6a710c0fbd78245340a2fce45aefbaa3da6dc/packages/nx/src/utils/cache-directory.ts#L37) nx cache dir changed to `.nx/cache` which by default is not gitignored.

### Expected Behavior

Expected behavior would be to have the nx cache directory gitignored by default.

### Current Behavior

Currently with recent changes nx cache directory is not gitignored.

### Reproduction Steps

Upgarding dependencies from `"nx": "^15.7.0",` to `"nx": "^15.8.5",` this behavior appeared.

### Possible Solution

add `/.nx/cache` to gitignore in `NxMonorepoProject` (or optionally in `PDKProject`)

_No response_

### PDK version used

0.14.10

Typescript

MacOS 13.2.1

agdimech:
This PR adds PNPM support to the PDK which results in huge performance gains.

```
pnpm i # installs all packages
pnpm build # alias for pnpx nx run-many --target=build --output-style=stream --nx-bail 
```

In addition to performance gains, due to the nature of how PNPM sets up node_modules there is now no longer a need to hoist anything as everything gets symlinked from child packages. This solves the #282 .

NXMonorepo has also been updated to add a gitignore for `.nx/cache` which solves #308 .

Sharp prebuild is also fixed as previously we were not including the vendor and build folders when bundling cdk-graph-plugin.

NOTE: I have added a jsii-pacmak hack for the time being until https://github.com/aws/jsii/pull/4030 is merged in order to prevent the build from breaking. Once that PR is merged and published, I will remove this workaround ASAP.

Fixes: #308, #282


agdimech:
This PR is dependent on https://github.com/aws/jsii/pull/4021

packages/open-api-gateway/package.json
```diff
@@ -36,7 +36,8 @@
   },
   "devDependencies": {
     "@aws-prototyping-sdk/nx-monorepo": "^0.x",
-    "@aws-sdk/client-s3": "^3.252.0",
+    "@aws-sdk/client-s3": "^3.6.1",
```
cogwirrel: Is there a reason for this downgrade? Wonder if it's amplify related as that uses 3.6.1 iirc.

The custom resource uses `transformToString` on the Body returned from an S3 `GetObjectCommand` which was introduced in 3.129.0: https://github.com/aws/aws-sdk-js-v3/blob/main/CHANGELOG.md#31290-2022-07-13

cogwirrel (commented):
Looks great - excited for the perf boost! Just one small comment :)

packages/open-api-gateway/package.json
agdimech: this isn't a downgrade, its an upgrade.

JeremyJonas: `^3.252.0` => `^3.6.1` is a "[downgrade](https://jubianchi.github.io/semver-check/#/^3.252.0/^3.6.1)"

.github/workflows/release.yml
```diff
@@ -30,14 +30,16 @@ jobs:
...
+        run: npm install -g pnpm
       - name: Install dependencies
-        run: yarn install --check-files --frozen-lockfile
+        run: pnpm i --frozen-lockfile
       - name: Synth Project
-        run: npx projen
+        run: pnpm projen
       - name: Build
-        run: yarn build
+        run: pnpm build --nx-bail
```
JeremyJonas: Pretty sure `--nx-bail` is already in the build task command.

.github/workflows/upgrade-mainline.yml
```diff
@@ -18,12 +18,14 @@ jobs:
...
+        run: npm install -g pnpm
       - name: Install dependencies
-        run: yarn install --check-files --frozen-lockfile
+        run: pnpm i --frozen-lockfile
       - name: Upgrade dependencies
-        run: npx projen upgrade-deps
+        run: pnpm projen upgrade-deps
       - name: Build
-        run: npx nx run-many --target=build --all --parallel=4 --output-style stream
+        run: pnpm build --output-style stream
```
JeremyJonas: nitpick: `--output-style stream` is the default, making this redundant

.gitignore
```diff
@@ -53,4 +54,11 @@ build
 .env
 .venv
 tsconfig.tsbuildinfo
+.yarn/unplugged
+.yarn/install-state.gz
+.yarn/cache
+.yarn/__virtual__
```
JeremyJonas, line 60: Should we also ignore the yarn.lock file since now using pnpm? Best to remove that file as well if not already as could be missleading and cause future issues.

.projen/tasks.json
```diff
@@ -121,9 +118,22 @@
...
+    "install:ci": {
+      "name": "install:ci",
+      "description": "Install project dependencies using frozen lockfile",
       "steps": [
         {
-          "exec": "npx nx run-many --target install-py --projects pipeline-sample-py --parallel=1"
+          "exec": "pnpm i --frozen-lockfile"
         }
       ]
     },
```
JeremyJonas: So in `install:ci` now we do not need to `install-py` within projects? How will py deps get installed during CI?

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -27,6 +27,23 @@ import { DEFAULT_CONFIG, SyncpackConfig } from "./syncpack-options";
...
+    case NodePackageManager.YARN:
+    case NodePackageManager.YARN2:
+      return "yarn";
+    case NodePackageManager.PNPM:
+      return "pnpx";
+    default:
+      return "npx";
+  }
+}
+
```
JeremyJonas: `execute` is a verb and expect this function to actually perform the execution, and would expect to be able to pass the command to execute as a parameter as well - consider renaming.

Would also be nice to have this take optional command that would get appended to clean up the implementations below: `${execute(this.package.packageManager)} foobar` to `execute(this.package.packageManager, "foobar")` (both of which result in `pnpx foobar`, etc.

Maybe we should add a public method on `NxMonorepo` for this, which would already have context of `this.package.packageManager` and support subclasses or child project to utilize this functionality.

```ts
class NxMonorepo {
 //...
 buildExecutableCommand(...command: string[]): string {
   return execute(this.package.packageManager) + " " + command.join(" ");
 }
}
```

private/projects/cdk-graph-plugin-diagram.ts
```diff
@@ -114,4 +117,16 @@ export class CdkGraphPluginDiagramProject extends CdkGraphPluginProject {
...
+    switch (this.package.packageManager) {
+      case NodePackageManager.YARN:
+      case NodePackageManager.YARN2:
+        return "yarn workspace";
+      case NodePackageManager.PNPM:
+        return "pnpm --filter";
+      default:
+        return "npx -p";
+    }
+  }
```
JeremyJonas: Similar comment as `execute` on NxMonorepo, this is verb and should be renamed. Additionally consider support taking in command to append for cleaner implementation usage.

Should we actually add this functionality to PDKProject so that all subclass, such as this, can utilize this common functionatlity? I would make it public on PDKProject as pretty core functionality that is beneficial.

private/projects/pdk-monorepo-project.ts
```diff
@@ -205,8 +196,16 @@ export class PDKMonorepoProject extends NxMonorepoProject {
    */
   synth() {
     this.subProjects.forEach((subProject) => {
+      subProject.packageTask.prependExec(
+        `pnpx ts-node ${path.relative(
+          subProject.outdir,
+          this.outdir
+        )}/scripts/jsii-pacmak-hack.ts`
+      );
```
JeremyJonas: Is the jsii-packman PR not published yet? Once that is this will get removed?

Can you add some comments regarding the clean up for this, and an issue to remove this once that PR is published.

JeremyJonas (commented):
This is really great work - massive performance and maintenance improvements!!!

JeremyJonas (approved):

PR #334, opened 2023-03-29
chore: upgrade pnpm v8 and implement "pdk-init" workflow action

JeremyJonas:
- Since v8 has breaking changes on the lock file, we need to ensure the engine and deps are using v8 throughout
- Standardizes how we init workflow jobs for engine, install, synth, tasks in a reusable local action.
  - adds pnpm setup
  - uses caching for dep

### DIFF
--- a/.github/workflows/aws-arch-pricing.yml
+++ b/.github/workflows/aws-arch-pricing.yml
@@ -16,10 +16,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: mainline
-      - name: Install pnpm
-        run: npm install -g pnpm
-      - name: Install dependencies
-        run: pnpm i --frozen-lockfile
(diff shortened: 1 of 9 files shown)


.github/workflows/aws-arch-pricing.yml
```diff
@@ -16,10 +16,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: mainline
-      - name: Install pnpm
-        run: npm install -g pnpm
-      - name: Install dependencies
-        run: pnpm i --frozen-lockfile
+      - name: PDK Init
+        uses: ./.github/actions/pdk-init
```
agdimech, line 20: Nice - didn't know you could refer to local actions!

agdimech (approved):

PR #337, opened 2023-04-10
fix: packageManager isn't configured properly

riywo:
The recent node.js returns "Usage error" if `packageManager` isn't
semver. This commit fixes the versions to semver complient.

Also, `commit-msg` hook is still using `yarn`. Since the package manager
for this repository is `pnpm`, this commit also updates the hook.

----

## Usage Error
```
❯ node -v
v18.15.0
❯ pnpm -v
8.2.0

# On the repository root
❯ pnpm -v
Usage Error: Invalid package manager specification in package.json; expected a semver version

$ pnpm ...
```

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn commitlint --edit
+pnpm commitlint --edit
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
   "license": "Apache-2.0",
   "version": "0.0.0",
   "types": "lib/index.d.ts",
-  "packageManager": "pnpm@8",
+  "packageManager": "pnpm@8.0.0",
   "private": true,
   "config": {
     "commitizen": {
--- a/packages/nx-monorepo/src/nx-monorepo.ts
+++ b/packages/nx-monorepo/src/nx-monorepo.ts
@@ -220,17 +220,17 @@ export class NxMonorepoProject extends TypeScriptProject {
         this.package.addEngine("pnpm", ">=8");
         // https://nodejs.org/api/corepack.html
         // https://nodejs.org/api/packages.html#packagemanager
-        this.package.addField("packageManager", "pnpm@8");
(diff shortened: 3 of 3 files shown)


JeremyJonas:
@riywo  That is interesting, the current value works for me doing the following as you described.

# On the repository root
❯ pnpm -v
Outputs: `8.0.0`

After upgrading from 8.0.0 to 8.2.0 locally it worked as well.
❯ pnpm -v
8.2.0

Does "packageManager" value support non-explicit version - like `pnpm@^8.0.0`? Would be ideal to not pin to explicit version so we don't need to maintain.

Are you using "corepack"? I added support but haven't tried it yet, curious if that is where the error is generating from.

JeremyJonas:
@riywo If we can't use non-explict version, might just be best to remove the "packageManager" value for now, since is experimental and for this project not being used atm. Unless you are using it?

If you're using`corepack` and this feature that is great, and we can pin like you have. Let know either way, and good to go for this.

riywo:
@JeremyJonas You mean `corepack`? Then, yes, I use `corepack` for `pnpm`.

It looks like this doesn't happen with locally installed `pnpm`:

```
❯ npm i pnpm

added 1 package, and audited 2 packages in 528ms

1 package is looking for funding
  run `npm fund` for details

found 0 vulnerabilities

❯ cat package.json
{
  "packageManager": "pnpm@8",
  "dependencies": {
    "pnpm": "^8.2.0"
  }
}

❯ npx pnpm -v
8.2.0
```

The version's description in the doc has nothing but it looks like pinned:

```
❯ cat package.json
{
  "packageManager": "pnpm@8.0.0"
}
❯ pnpm -v
8.0.0

❯ cd ~
❯ pnpm -v
8.2.0
```

I would suggest not to opt-in `packageManager` now since it's experimental and only supports pinned version. Either getting rid of it or pinning the version is the only way to support `corepack` users at this moment.

JeremyJonas:
@riywo Ok lets remove `packageManager` from package.json, that will be clearest fix. Thank you!

agdimech:
Closing this PR given it has been implemented in: #339. Feel free to re-open if you feel this was closed in error.

PR #338, opened 2023-04-10
fix(pipeline): remove unneeded cfn exports

### LINKED ISSUE #239: [BUG] Only one PDKPipeline can be created per account/region
### Describe the bug

See: https://github.com/aws/aws-prototyping-sdk/discussions/235

The `exportName` for the `CodeRepositoryGRCUrl` is hardcoded and so conflicts with a second deployment of `PDKPipeline`.

### Expected Behavior

Able to deploy multiple `PDKPipeline` constructs in the same account/region

### Current Behavior

```
Export with name CodeRepositoryGRCUrl is already exported by stack PipelineStack. Rollback requested by user.
```

### Reproduction Steps

Deploy more than one `PDKPipeline`, either in the same app or as separate apps.

### Possible Solution

Include `props.repositoryName` in the export name since this is already a unique name.

_No response_

### PDK version used

0.12.4

Typescript, Java, Python

OSX

jstrunk:
The pipeline package was exporting `CodeRepositoryGRCUrl` and `SonarqubeSecretArn` which meant the stack could only be deployed once per region. Nothing seems to be importing these exports. After discussing with the maintainers, we determined that simple outputs are sufficient.

BREAKING CHANGE: `CodeRepositoryGRCUrl` and `SonarqubeSecretArn` are no longer CloudFormation exports. Any systems that relied on those being exports will need to be updated to look up the output value by the stack name.

Fixes #239


agdimech:
Thanks for submitting this!

agdimech (approved):

PR #339, opened 2023-04-11
feat: pnpm bundling

JeremyJonas:
- Adds transitive bundled dependencies symlinks to packages during package task
- Use `NX_NON_NATIVE_HASHER=true` for all run-many tasks
- Upgrade nx versions
- Use `--runInBand` for nx run-many jest executions


packages/nx-monorepo/scripts/pnpm/link-bundled-transitive-deps.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (lockfile == null) {
+      throw new Error("Failed to read lockfile")
+    }
+  }
+
+  const pkgDeps = lockfile.importers[pkgFolder].dependencies!;
+
+  const transitiveDeps: Record<string, { version: string, path: string }> = {};
+
+  function visit(_deps?: ResolvedDependencies) {
```
agdimech: Why is this optional? If it needs to remain optional then we should change line 36 to: if `(!deps || _deps == null)`

JeremyJonas: The conditional to check if undefined, null, or empty is inside the function, rather than have to check before each call.

.husky/pre-commit
```diff
@@ -5,9 +5,9 @@ set -eo pipefail
 
 PRE_SHA="$(git diff HEAD --ignore-space-at-eol | shasum -a 256 --tag)"
```
agdimech, line 6: We should also update commit-msg to use pnpm

private/projects/pdk-monorepo-project.ts
```diff
@@ -175,9 +172,15 @@ export class PDKMonorepoProject extends NxMonorepoProject {
       ".pnp.loader.cjs"
     );
 
+    const npmrc = new IniFile(this, ".npmrc", {});
     // add local `.npmrc` to automatically avoid build hangs if npx is promping to install a package
-    const npmrc = new NpmConfig(this);
-    npmrc.addConfig("yes", "true");
+    npmrc.addOverride("yes", true);
+    // https://pnpm.io/npmrc#public-hoist-pattern
+    npmrc.addOverride("public-hoist-pattern", [
```
agdimech: Why do we need to hoist these?

private/projects/nx-monorepo-project.ts
```diff
@@ -17,16 +17,21 @@ export class NXMonorepoProject extends PDKProject {
...
+      bundledDeps: ["@nrwl/devkit", "fs-extra", "semver"],
       stability: Stability.STABLE,
     });
 
     this.compileTask.exec(
       'rsync -a ./src/** ./lib --include="*/" --include="**/*.js" --exclude="*" --prune-empty-dirs'
     );
 
-    this.package.addBin({ "nx-dir-hasher": "./scripts/nx-dir-hasher.js" });
+    this.package.addBin({ "pdk@nx-dir-hasher": "./scripts/nx-dir-hasher.js" });
```
agdimech, line 35: any reason why we are prefixing this with `pdk@`?

private/projects/pdk-monorepo-project.ts
JeremyJonas: eslint and prettier are the defaults for pnpm `public-hoist-pattern`, so need append `@pnpm/*` also needed to add these back. The `@pnpm/*` was needed when I used the pnpm lock instead of pnpm list to get deps and path - but not needed now.

I will remove this chunk - not needed now.

.husky/pre-commit
JeremyJonas: Ack

packages/nx-monorepo/scripts/pnpm/link-bundled-transitive-deps.ts
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env ts-node
+/*! Copyright [Amazon.com](http://amazon.com/), Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0 */
+import * as path from "node:path";
+import * as child_process from "node:child_process";
+import { promisify } from "node:util";
+import * as fs from "fs-extra";
+import * as SemVer from "semver";
+
+interface Dependency {
```
agdimech: Can you just use: https://github.com/pnpm/pnpm/blob/main/reviewing/list/src/getPkgInfo.ts#L13 ?

.husky/pre-commit
JeremyJonas: Done

private/projects/nx-monorepo-project.ts
JeremyJonas: Since these are bins the are available in PATH, and want to have clear scope for PDK based bin commands so not confused with other `nx-*` commands, and similarly with `pnpm-*` commands. The `@` is supported in filename cross-env so provides nice namespace separator.

packages/nx-monorepo/scripts/pnpm/link-bundled-transitive-deps.ts
JeremyJonas: Nice! Actually just found https://github.com/pnpm/pnpm/blob/main/reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts#L31, which is exactly what we need - no need to call `pnpm list` - we can just use it all programmatically for exactly what we need 😄

JeremyJonas: Done

packages/nx-monorepo/scripts/pnpm/link-bundled-transitive-deps.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as path from "node:path";
+import * as fs from "fs-extra";
+import * as SemVer from "semver";
+import { buildDependenciesHierarchy, PackageNode } from "@pnpm/reviewing.dependencies-hierarchy";
+
+async function linkBundledTransitiveDeps(workspaceDir: string, pkgFolder: string) {
+  const pkgDir = path.join(workspaceDir, pkgFolder);
+  const pkgJson = require(path.join(pkgDir, "package.json"));
+  const bundledDeps: string[] = pkgJson.bundledDependencies || [];
+  if (bundledDeps.length < 1) {
```
agdimech: NIT `bundledDeps.length == 0` ?

packages/nx-monorepo/scripts/pnpm/link-bundled-transitive-deps.ts
```diff
@@ -0,0 +1,90 @@
...
+      optionalDependencies: false,
+      dependencies: true,
+      devDependencies: false,
+    },
+  }))[pkgDir];
+
+  const transitiveDeps: Record<string, PackageNode> = {};
+
+  function visit(_deps?: PackageNode[]) {
+    if (_deps == null || !_deps.length) {
```
agdimech, line 31: deps is optional so we also need to check for `!_deps` before checking `!_deps.length`

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -534,6 +564,64 @@ export class NxMonorepoProject extends TypeScriptProject {
...
+  preSynthesize(): void {
+    super.preSynthesize();
+
+    if (this._options.workspaceConfig?.linkLocalWorkspaceBins === true) {
+      this.linkLocalWorkspaceBins();
+    }
+
+    if (this.package.packageManager === NodePackageManager.PNPM) {
+      // PNPM hoisting hides transitive bundled dependencies which results in
+      // transitive dependencies being packed correctly.
```
agdimech: packed incorrectly*

packages/nx-monorepo/scripts/pnpm/link-bundled-transitive-deps.ts
JeremyJonas: That is what `_deps == null` is doing

agdimech: _deps == null is not the same as _deps === undefined right?

JeremyJonas: `x == null` is the same as `x === null || x === undefined`. The reason linters allow loose equality for `== null` use case is specifically to allow for this.

JeremyJonas: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#difference_between_null_and_undefined

agdimech (approved):

PR #341, opened 2023-04-12
fix: move implicitDependencies to individual package.json to reduce c…

### LINKED ISSUE #336: [FEATURE] Allow `addImplicitDependency` by string
### Describe the feature

Same as #335, but more specific to one use case i.e. adding implicit dependency just by strings, not by projects.

### Use Case

Adding a codegen package to workspace so that local dependencies can be resolved properly without extra effort. Then, this feature allows us to specify those dependencies by just using package names.

### Proposed Solution

Accept `string` as parameters of `addImplicitDependency`.

```ts
  public addImplicitDependency(
    dependent: Project | string,
    dependee: Project | string
  ) {
    const dependentName =
      dependent instanceof Project ? dependent.name : dependent;
    const dependeeName = dependee instanceof Project ? dependee.name : dependee;
    if (this.implicitDependencies[dependentName]) {
      this.implicitDependencies[dependentName].push(dependeeName);
    } else {
      this.implicitDependencies[dependentName] = [dependeeName];
    }
  }
```

### Other Information

_No response_

### Acknowledgements

### PDK version used

0.14.21

Typescript

Linux 0eac0dfd53e7 5.15.0-69-generic #76-Ubuntu SMP Fri Mar 17 17:25:19 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux

agdimech:
This PR simplifies the way implicitDependencies are handled by moving them to the specific package.json that has the dependency.

This removes the need for a custom NX plugin which was needed due to the fact that it only renders implicit dependencies that are present on the package itself.

Fixes: #336


packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -567,6 +556,17 @@ export class NxMonorepoProject extends TypeScriptProject {
...
+      const manifest = this.subProjects
+        .find((p) => p.name === packageName)
+        ?.tryFindObjectFile("package.json");
+
+      if (manifest) {
+        manifest.addOverride("nx.implicitDependencies", deps);
+        manifest.synthesize();
+      }
+    });
   }
```
JeremyJonas: Should also support `project.json` for non-NodeProject support.

Can `addImplicitDependency` just add the dep directly to the project rather than store it in `this.implicitDependencies`? I think we should create an `NxProjectConfig` Component (Projen Component), and the PDKProject will already have it added, but for other project types that don't have the component added, the NxMonorepo can add it. 

The NxProjectConfig component could then handle all the config specific stuff we need for Nx... then during synth, it would either modify the `package.json` for NodeProjects, or create a `project.json` file for "non-NodeProjects".

Much more composable way that scales to support non PDKProject stuff as well.

`NxProjectConfig.of(dependent).addImplicitDependency(dependee)` would be great. The `.of` method would create the NxProjectConfig for project if doesn't already have.

agdimech: Updated in latest revision.

packages/aws-arch/package.json
```diff
@@ -203,7 +203,8 @@
           "{projectRoot}/src/generated"
         ]
       }
-    }
+    },
+    "implicitDependencies": []
```
JeremyJonas: Nitpick: remove property if empty

packages/cdk-graph/package.json
```diff
@@ -189,5 +189,8 @@
       "@types/yargs": "17.0.10"
     }
   },
+  "nx": {
+    "implicitDependencies": []
+  },
```
JeremyJonas: Nitpick: remove entire `nx` block if empty

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -491,12 +486,8 @@ export class NxMonorepoProject extends TypeScriptProject {
...
    * @param dependee project you wish to depend on.
    */
-  public addImplicitDependency(dependent: Project, dependee: Project) {
-    if (this.implicitDependencies[dependent.name]) {
-      this.implicitDependencies[dependent.name].push(dependee.name);
-    } else {
-      this.implicitDependencies[dependent.name] = [dependee.name];
-    }
+  public addImplicitDependency(dependent: Project, dependee: Project | string) {
+    NxProject.of(dependent)?.addImplicitDependency(dependee);
```
JeremyJonas: Should we throw error if `dependent` doesn't have `NxProject`? Otherwise it is a noop without warning that the implicit dependency wasn't set? I know we are adding NxProject to subProjects above, but to would be nice to future proof.

packages/nx-monorepo/src/nx-project.ts
```diff
@@ -0,0 +1,69 @@
+/*! Copyright [Amazon.com](http://amazon.com/), Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0 */
+import * as path from "path";
+import { Component, JsonFile, Project } from "projen";
+import { NodeProject } from "projen/lib/javascript";
+import { buildExecutableCommand } from "./nx-monorepo";
+
+export class NxProject extends Component {
```
JeremyJonas: Needs docs throughout this file for all exports.

packages/nx-monorepo/src/nx-project.ts
```diff
@@ -0,0 +1,69 @@
...
+  static of(project: Project): NxProject {
+    return project.components.find((c) => c instanceof NxProject) as NxProject;
+  }
+
+  private readonly implicitDependencies: string[] = [];
+  constructor(project: Project) {
```
JeremyJonas: Linebreak

packages/nx-monorepo/src/nx-project.ts
```diff
@@ -0,0 +1,69 @@
...
+    super(project);
+  }
+
+  public addImplicitDependency(dependee: Project | string) {
+    this.implicitDependencies.push(
+      dependee instanceof Project ? dependee.name : dependee
+    );
+  }
```
JeremyJonas: Nice - will fix #336

packages/nx-monorepo/src/nx-project.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  synthesize() {
+    if (this.project instanceof NodeProject) {
+      this.project
+        .tryFindObjectFile("package.json")
+        ?.addOverride("nx.implicitDependencies", this.implicitDependencies);
```
JeremyJonas: Should this use [addToArray](https://github.com/projen/projen/blob/main/src/object-file.ts#L156), which would support existing values of `nx.implicitDependencies` being added prior to this call.

packages/pipeline/samples/java/project.json
```diff
@@ -0,0 +1,73 @@
...
+  "name": "pipeline-sample-java",
+  "root": "packages/pipeline/samples/java",
+  "implicitDependencies": [
+    "aws-prototyping-sdk"
+  ],
+  "targets": {
+    "clobber": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "pnpx  projen clobber",
```
JeremyJonas: Nitpick: extra space after "pnpx" in following commands: `pnpx  projen clobber`

packages/nx-monorepo/src/nx-project.ts
```diff
@@ -0,0 +1,69 @@
...
+      projectJson.addOverride(
+        "targets",
+        this.project.tasks.all.reduce(
+          (p, c) => ({
+            [c.name]: {
+              executor: "nx:run-commands",
+              options: {
+                command: `${buildExecutableCommand(
+                  (this.project.root as NodeProject).package.packageManager
+                )} projen ${c.name}`,
```
JeremyJonas: `buildExecutableCommand` already has extra space... maybe we should fix it there - and only have space when args are provided?

packages/aws-arch/package.json
agdimech: Updated.

packages/cdk-graph/package.json
agdimech: Updated.

packages/nx-monorepo/src/nx-monorepo.ts
agdimech: Updated.

packages/nx-monorepo/src/nx-project.ts
agdimech: Updated.

packages/nx-monorepo/src/nx-project.ts
agdimech: Fixed.

packages/nx-monorepo/src/nx-project.ts
agdimech: :)

packages/nx-monorepo/src/nx-project.ts
agdimech: Would prefer to keep as is as otherwise I will need NodeProject vs Non-NodeProject branching in two places opposed to one.

packages/pipeline/samples/java/project.json
JeremyJonas: Don't worry about this one - I have fixed in my PR

packages/nx-monorepo/src/nx-project.ts
agdimech: Fixed.

packages/pipeline/samples/java/project.json
agdimech: Sorry - already fixed :)

packages/nx-monorepo/src/nx-project.ts
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * @param project project instance.
+   */
+  static of(project: Project): NxProject | undefined {
+    return project.components.find((c) => c instanceof NxProject) as
+      | NxProject
+      | undefined;
+  }
+
+  private readonly implicitDependencies: string[] = [];
```
JeremyJonas: Nitpick: Private should be `_` prefix and `@internal`.

packages/nx-monorepo/src/nx-project.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  constructor(project: Project) {
+    super(project);
+  }
+
+  /**
+   * Adds a implicitDependency between the dependant (this project) and dependee.
```
JeremyJonas: Nitpick: "Adds an implicit dependency..."
- "an"
- "implicit dependency"

packages/nx-monorepo/src/nx-project.ts
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * @param dependee project to add the implicitDependency on.
```
JeremyJonas: "implicit dependency"

JeremyJonas (approved):
Awesome work @agdimech 🚀 , really nice improvements. 

Just a few minor nitpicks - but still approved!

PR #344, opened 2023-04-13
feat(type-safe-api): add type safe api, successor to open api gateway

### LINKED ISSUE #192: [FEATURE] Make Smithy Build/Model a first class package
### Describe the feature

Can we pull Smithy Build and Model out of OpenAPIGateway and make it it's own project?

### Use Case

Ive been using SmithyAPIGatewayTSProject to generate typescript API. Instead of generating code from an OpenAPI spec, I wanted to generate code from Smithy directly instead. The current solution has been:

- Just generate and ignore/delete the generated code for OpenAPI
- Use the generated SDKs instead

I keep running into a chicken and egg problem:

- Opinionated CDK Construct uses generated SDKs
- SDKs are generated within the same project

nx tries to find and build all subdirectories with a package.json. So what ends up happening is that the CDK construct tries to build before the model builds. You can get around this with some projen-fu by generated SDKs, manually running `yarn install` to install the net new dependencies the SDK needs, commenting out files, adding dependencies, uncommenting files, etc, etc etc.

It would be easier if the flow was:

```bash
packages/
|--model        # contains model and generated code
|--construct   # contains CDK constructs that depends on model
|--anything else down the road # depends on model
``` 

This way, all packages can depend on model, that way, whenever I run `npx nx run-many --target build --all`. `nx` will know to always build model first.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### PDK version used

0.12.4

_No response_

MacOS

cogwirrel:
This change introduces Type Safe API, which replaces Open API Gateway going forward. We rename since with Smithy support, having "Open API" as part of the name is less relevant to users. Type Safe API emphasises one of the main benefits of the package, type-safety.

The fundamental difference in this new package is that code generation is now a build task, rather than coupled to project synthesis. This eliminates the need to check generated code in to version control, meaning smaller and more legible code reviews.

The project is now more modular, with a cleaner separation of the model, infrastructure, clients and documentation, resulting in a more intuitive directory structure.

Instead of the 6 different projen projects vended by open-api-gateway (which were specific to the model language (Smithy vs OpenAPI) and infrastructure language (TS, Java or Python), this vends a single `TypeSafeApiProject` which lets you configure the infrastructure and model language. Additionally, we expose the `TypeSafeApiModelProject` to allow for  the Smithy/OpenAPI model project to be used on its own, which addresses #192.

We remove the opinionated "SampleApi" generated for each language, since while this was good for enabling users to deploy a hello world API with very little code change required, it involved starting users off in a less ideal way. In Java the whole CDK project was built as a superjar and deployed as a lambda (much larger jar than you'd really want). In Python the API client was built as a layer rather than using local dependencies. In typescript we put the handlers in the api infra project and relied on `NodejsFunction`'s behaviour of looking for both `.ts` and `.js` files so it'd work as a library or the root cdk app. So instead of setting users up with a less optimal project structure which "just works" out of the box, we include quick start guides in the README for how to set up different languages in the right way.

Note that the README can be further extended and broken up into separate files, but should be a good start for now.

Fixes #192

Other changes:
* Upgraded projen since I needed [this fix](https://github.com/projen/projen/pull/2602) - this overrode our overridden eslint tasks so I added a fix in `nx-monorepo.ts` to make sure we retain our eslint tasks
* Disabled the `open-api-gateway` tests since that package is deprecated and it's the biggest contributor to our build time by a large margin! Tests for `type-safe-api` are much speedier :)


packages/type-safe-api/README.md
```diff
@@ -0,0 +1,1423 @@
...
+}
+```
+
+#### Supported Protocols
+
+Currently only [AWS restJson1](https://smithy.io/2.0/aws/protocols/aws-restjson1-protocol.html) is supported. Please ensure your service is annotated with the `@restJson1` trait.
+
+#### Multiple Files
+
+You can split your definition into multiple files and folders, so long as they are all under the `smithy/src/main/smithy` directory in your API project.
```
sperka: ~smithy~`model/src/main/smithy`

packages/open-api-gateway/README.md
```diff
@@ -1,3 +1,33 @@
...
+
+Please use the [`type-safe-api`](../type-safe-api) package instead. `open-api-gateway` will be removed in the GA 1.0 release.
+
+`type-safe-api` offers several benefits over `open-api-gateway`:
+
+* Code generation at build time:
+  * You no longer need to run `npx projen` when you make changes to your API definition.
+  * You no longer need to pollute your version control with generated code
+* More intuitive directory structure
+* Generated python projects use `poetry`, allowing for better local dependencies (eg no need to deploy a lambda layer to for handlers to depend on the generated handler wrappers)
```
agdimech: lambda layer for*

packages/open-api-gateway/README.md
```diff
@@ -1,3 +1,33 @@
...
+
+### TypeSafeApi Migration Guide
+
+Migration will vary a little depending on the language you are using, however the following rough steps can be used to migrate a project use `type-safe-api`:
```
agdimech: a project to use*

packages/open-api-gateway/README.md
```diff
@@ -1,3 +1,33 @@
...
+1. Duplicate any client dependencies in your `.projenrc.ts` so that packages depend on the equivalent package. (eg. if a package depends on `api.generatedTypescriptClient`, make sure it depends on `newApi.typescriptClient`).
+1. Duplicate any infrastructure dependencies in your `.projenrc.ts` - the infrastructure is now a separate subproject rather than the top level API project. (eg if a package depended on `api` before, it should now depend on `newApi.typescriptInfrastructure`)
+1. Synthesize with `npx projen`
+1. Copy your model into the new directory:
+   1. For Smithy, you can copy everything in `<oldoutdir>/smithy/src/main/smithy` directory into `<newoutdir>/model/src/main/smithy`
+   1. For OpenAPI, copy your `<oldoutdir>/src/spec.yaml` to `<newoutdir>/model/src/main/openapi`, and rename it to `main.yaml`. Additionally, copy any other referenced OpenAPI files from `<oldoutdir>/src/` into `<newoutdir>/model/src/main/openapi` if present.
+1. Run a build (eg. `npx nx run-many --target build --all` in root of monorepo if using monorepo).
```
agdimech: this can be simplified to `npm/yarn/pnpm build`

private/projects/type-safe-api-project.ts
```diff
@@ -0,0 +1,67 @@
...
+        "@types/lodash",
+        "aws-cdk-lib",
+        "cdk-nag",
+        "constructs",
+        "projen",
+        "@aws-sdk/client-s3",
+      ],
+      deps: ["@aws-prototyping-sdk/pdk-nag@^0.x", "fs-extra"],
+      bundledDeps: [
+        'date-format',  'debug',
```
agdimech: newline per dep

agdimech: Also: are these all required now that @JeremyJonas fix has been added?

packages/type-safe-api/README.md
```diff
@@ -0,0 +1,1423 @@
...
+  }
+}
+```
+
+#### Implement a Lambda Handler
+
+The generated client projects include lambda handler wrappers which provide type-safety for implementing your API operations. You can implement your lambda handlers in any of the supported languages, and even mix and match languages for different operations if you like. In typescript, the implementation of `say-hello.ts` would look like:
+
+```ts
+import { sayHelloHandler } from "myapi-typescript-client"; // <- generated typescript client package
```
agdimech: DISCUSSION:

Are we grouping generated clients with generated handlers? Seems a little unintuitive to have this belong in the client package. Perhaps seperating this out into a client vs server may be easier to understand?

packages/type-safe-api/README.md
```diff
@@ -0,0 +1,1423 @@
...
+
+The `TypeSafeApiProject` will create the following directory structure within its `outdir`:
+
+```
+|_ model/
+    |_ src/
+        |_ main/
+            |_ smithy - your API definition if you chose ModelLanguage.SMITHY
+            |_ openapi - your API definition if you chose ModelLanguage.OPENAPI
+|_ clients/ - generated clients in the languages you specified
```
agdimech: I assume this encompasses actual clients and server handlers?

packages/type-safe-api/README.md
```diff
@@ -0,0 +1,1423 @@
...
+    |_ markdown
+```
+
+
+### Smithy IDL
+
+Please refer to the [Smithy documentation](https://smithy.io/2.0/quickstart.html) for how to write models in Smithy. A basic example is provided below:
+
+```smithy
+$version: "2"
```
agdimech, line 257: Is this the generated sample by default?

packages/type-safe-api/README.md
```diff
@@ -0,0 +1,1423 @@
...
+properties:
+  message:
+    type: string
+required:
+  - message
+```
+
+### Construct
+
+A sample construct is generated which provides a type-safe interface for creating an API Gateway API based on your OpenAPI specification. You'll get a type error if you forget to define an integration for an operation defined in your api.
```
agdimech: I thought we weren't generating this anymore? If we are, I assume this is controlled via a sample flag on the projen construct?

packages/type-safe-api/README.md
```diff
@@ -0,0 +1,1423 @@
...
+        },
+      },
+    });
+  }
+}
+```
+
+#### Sharing Integrations
+
+If you would like to use the same integration for every operation, you can use the `Operations.all` method from your generated client to save repeating yourself, for example:
```
agdimech: We should explain why you would want to do this. Perhaps link to below handlerRouter section?

packages/type-safe-api/samples/smithy/gradlew.bat
```diff
@@ -0,0 +1,91 @@
+@rem
```
agdimech, line 1: Why are all the smithy files here apart of samples? Shouldn't there be a sample smithy IDL here similar to how you have spec.yaml?

packages/type-safe-api/scripts/common/common.sh
```diff
@@ -0,0 +1,66 @@
...
+}
+
+##
+# runs the passed command with the package manager's proper syntax
+run_command() {
+  cmd="$@"
+
+  if [ "$pkg_manager" == "pnpm" ]; then
+    # pnpx cli was deprecated in v6 --> use pnpm dlx
+    runner="pnpm dlx"
```
agdimech, line 54: should probably support yarn as well. Can add a issue for that as its not needed right now.

packages/type-safe-api/src/project/type-safe-api-project.ts
```diff
@@ -0,0 +1,287 @@
...
+  generateInfraProject,
+} from "./codegen/generate";
+import { DocumentationFormat, Language } from "./languages";
+import { TypeSafeApiModelProject } from "./model/type-safe-api-model-project";
+import { GeneratedCodeOptions, ModelLanguage, ModelOptions } from "./types";
+
+/**
+ * Options for the TypeSafeApiProject
+ */
+export interface TypeSafeApiProjectOptions extends ProjectOptions {
```
agdimech: very clean :)

private/projects/type-safe-api-project.ts
cogwirrel: Ooh good call I should be able to revert back to just the direct dependencies :)

packages/type-safe-api/README.md
cogwirrel: That's a good point! Separating them out would definitely be cleaner. I think it's possible - the tricky bit would be passing the right package names to OpenAPI generator so that we can still use the mustache templates, but import the models from the client package rather than relative paths.

I guess the big pro of this would be separation and you could even switch off generating the handler wrappers, but the con is it'd add another package to build which would slow dev time down a little bit... Would potentially open up the possibility to have eg `typescript-server-lambda` and `typescript-server-express` in future without bloating the clients too much though which could be useful!

packages/type-safe-api/README.md
cogwirrel: Yes - server handlers are generated as part of each client package

packages/type-safe-api/README.md
cogwirrel: Yeah :)

packages/type-safe-api/README.md
cogwirrel: Good call will reword - we generate the type safe construct but no sample which instantiates/extends it :)

packages/type-safe-api/README.md
cogwirrel: Cool will do!

packages/type-safe-api/samples/smithy/gradlew.bat
cogwirrel: Actually I can delete `spec.yaml` from here as it's no longer used :) For both Smithy and OpenAPI we include the `serviceName` or `title` specified in `modelOptions` in the sample model (ie it's defined in code not a static file).

eg: https://github.com/aws/aws-prototyping-sdk/blob/feat/type-safe-api/packages/type-safe-api/src/project/model/openapi/open-api-definition.ts#L31

packages/type-safe-api/scripts/common/common.sh
cogwirrel: Can add a separate issue, though I'm not sure it's very helpful as it's just used behind the scenes and yarn doesn't have a significant performance increase over npm :)

packages/type-safe-api/README.md
cogwirrel: As discussed offline, we'll rename this to `types` which will encompass both client and server side code. If in future we want to implement different client/server types/wrappers we can add configuration options to let users customise what they want, so as not to bloat the codebase :) This'll be the best of both worlds where the naming is more intuitive, it's still extensible, and users don't take the build time hit of having another package to build :)

agdimech (approved):
Great job Jack - this is a huge improvement!

PR #345, opened 2023-04-14
fix: use package manager specific exec commands

agdimech:
Currently the projen command is defaulted to `npx projen`. This PR uses the configured package manager to execute the command using locally installed packaged (instead of downloading and running).

This also improves the installation of non-node package dependencies by invoking the install target as a postinstall.


packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -721,75 +750,34 @@ export class NxMonorepoProject extends TypeScriptProject {
...
+      this.tasks.tryFind("postinstall") ?? this.addTask("postinstall");
     monorepoInstallTask.exec(
-      `${buildExecutableCommand(
-        this.package.packageManager
-      )} nx run-many --target install-py --projects ${pythonProjects
-        .map((project) => project.name)
-        .join(",")} --parallel=1`
+      buildExecutableCommand(
+        this.package.packageManager,
+        `nx run-many --target install --projects ${installProjects
```
cogwirrel, line 773: Much cleaner, nice one!

We should remove this too: https://github.com/aws/aws-prototyping-sdk/blob/mainline/packages/open-api-gateway/src/project/codegen/generated-python-client-project.ts#L65-L75

It's there because open-api-gateway synths stuff early, before the monorepo can make changes to its tasks... No such hack in type-safe-api though haha

JeremyJonas (approved):

packages/nx-monorepo/src/nx-monorepo.ts
agdimech: Updated. Good call

cogwirrel (approved):

PR #349, opened 2023-04-17
build: upgrade node version when release to package manager

agdimech:
Release to package managers is currently failing as pnpm requires NODE >16.


martenpayne (approved):
👍 looks good

PR #354, opened 2023-04-19
feat(identity): support updated MFA props and UserPoolWithMFA construct

agdimech:
- Refactor out UserPool into re-usable UserPoolWithMFA construct
- Support updated MFA props
- Fix caching issue with docs
- Support backwards compatibility with a `@aws-prototyping-sdk/identity:useLegacyMFAProps` feature flag

BREAKING CHANGE: UserPoolWithMfa construct has recently been added which provides a more well rounded set of default configurations for the UserPool. It introduces a breaking change in that existing customers of the UserIdentity construct prior to v0.15.5 will by default be upgraded to use the new defaults which will cause deployments to fail. To workaround this, a new cdk context property ‘@aws-prototyping-sdk/identity:useLegacyMFAProps’ was added to revert back to legacy defaults.

### DIFF
--- a/packages/identity/src/user-identity.ts
+++ b/packages/identity/src/user-identity.ts
@@ -5,16 +5,11 @@ import {
   IdentityPoolProps,
   UserPoolAuthenticationProvider,
 } from "@aws-cdk/aws-cognito-identitypool-alpha";
-import { PDKNag } from "@aws-prototyping-sdk/pdk-nag";
-import { Stack, Duration } from "aws-cdk-lib";
-import {
-  AccountRecovery,
-  CfnUserPool,
-  Mfa,
-  UserPool,
-  UserPoolClient,
-} from "aws-cdk-lib/aws-cognito";
+import { UserPool, UserPoolClient } from "aws-cdk-lib/aws-cognito";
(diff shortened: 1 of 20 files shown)


cogwirrel (approved):

packages/identity/src/user-identity.ts
```diff
@@ -45,74 +38,27 @@ export class UserIdentity extends Construct {
...
-          ...(!props?.userPool
-            ? [
-                new UserPoolAuthenticationProvider({
-                  userPool: this.userPool,
-                  userPoolClient: this.userPoolClient!,
-                }),
-              ]
-            : []),
-        ],
+    this.userPoolClient = this.userPool.addClient("WebClient", {
```
JeremyJonas: What if user pool is supplied in options and already contains "WebClient" client? Does UserPool support duplicate client names? Can we check if "WebClient" client already exists, and either throw an error or use the existing client?

JeremyJonas (approved):

packages/identity/src/user-identity.ts
agdimech: Updated.

PR #358, opened 2023-04-20
feat(nx-monorepo): simplify adding homogenous dependencies for java and python

### LINKED ISSUE #346: [GA] Add test to ensure nx-monorepo synth order is preserved
### Describe the bug

NX monorepo synths projects in alphabetical order due to the `subProjects` getter sorting `this.subprojects` in place with `.sort()`.

Noticed this while testing out https://github.com/aws/aws-prototyping-sdk/pull/344 and will fix as part of that PR, but opening an issue to make sure we add a test for this before GA.

### Expected Behavior

.

### Current Behavior

.

### Reproduction Steps

.

### Possible Solution

Write a test

_No response_

### PDK version used

0.15.0

Typescript, Java, Python

osx

cogwirrel:
Added two helper methods to the monorepo to make it easier to set up dependencies between python or java projects.

Also add a test to make sure the monorepo synthesizes subprojects in instantiation order, rather than alphabetical order (this was fixed in #344 but I raised #346 as a reminder to add a test)

Fixes #346


packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -599,6 +601,60 @@ export class NxMonorepoProject extends TypeScriptProject {
...
+    // Add implicit dependency for build order
+    this.addImplicitDependency(dependent, dependee);
+
+    // Add dependency in pom.xml
+    dependent.addDependency(
+      `${dependee.pom.groupId}/${dependee.pom.artifactId}@${dependee.pom.version}`
+    );
+
+    // Add a repository so that the dependency in the pom can be resolved
+    dependent.pom.addRepository({
```
agdimech, line 619: This will be OK even though the dep may not have dist/java? I assume this only matters as part of build as java doesn't have an install task?

cogwirrel: Good callout! Yeah maven would fail at build time. From some more digging it turns out maven packaging is always on for JavaProjects, but you can customise this dist dir. Sadly it's not exposed at the moment so I've raised this PR: https://github.com/projen/projen/pull/2624

Happy to wait until that gets merged and update, or to just assume dist/java and revisit when the dist dir is accessible :)

agdimech: Let's wait and fix it all at once :)

Also - your linked PR is failing as I assume you didn't do a full rebuild of projen.

agdimech (approved):

packages/nx-monorepo/src/nx-monorepo.ts
cogwirrel: As discussed offline, we'll assume `dist/java` for now but update when the projen PR is merged. Tracking here: https://github.com/aws/aws-prototyping-sdk/issues/362

PR #361, opened 2023-04-21
fix: make ACL's private

agdimech:
- Fixes https://aws.amazon.com/about-aws/whats-new/2023/04/amazon-s3-two-security-best-practices-buckets-default/
- Adds commitizen to root as a devDep

BREAKING CHANGE: Pipeline has been refactored into a construct in order to resolve the S3 issue. Accessing CodePipeline methods now require accessing the public pipeline member of PDKPipeline i.e. `pdkPipeline.codePipeline.XXX`

### DIFF
--- a/packages/pipeline/src/pdk-pipeline.ts
+++ b/packages/pipeline/src/pdk-pipeline.ts
@@ -1,7 +1,6 @@
 /*! Copyright [Amazon.com](http://amazon.com/), Inc. or its affiliates. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0 */
-import { PDKNag } from "@aws-prototyping-sdk/pdk-nag";
-import { Aspects, CfnOutput, RemovalPolicy, Stack, Stage } from "aws-cdk-lib";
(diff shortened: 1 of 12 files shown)


packages/pipeline/src/pdk-pipeline.ts
```diff
@@ -105,6 +106,7 @@ export class PDKPipeline extends CodePipeline {
       autoDeleteObjects: true,
       removalPolicy: RemovalPolicy.DESTROY,
       encryption: BucketEncryption.S3_MANAGED,
+      accessControl: BucketAccessControl.LOG_DELIVERY_WRITE,
```
cogwirrel: Not sure if it's worth using IAM and the bucket policy for the access logs bucket instead? (ref: https://github.com/aws/aws-cdk/issues/22183) I think this also requires a cdk.context.json feature flag too though

cogwirrel (approved):

packages/pipeline/src/pdk-pipeline.ts
agdimech: Fixed :)

PR #363, opened 2023-04-21
feat(type-safe-api): context option for synthesizing to sam local compatible template

### LINKED ISSUE #317: [FEATURE] Ability to test locally for `open-api-gateway`
### Describe the feature

When creating an API with `open-api-gateway`, the ability to test the generated API locally would be great. Currently a CDK deploy is needed to play around with the API.

### Use Case

Playing around/investigating things with the generated API on my local machine 

### Proposed Solution

AWS SAM has an option to run an API locally: `sam local start-api`, perhaps that could be used?

### Other Information

@cogwirrel mentioned:

> open-api-gateway points the API Gateway API at an OpenAPI specification that's written to S3 by a custom resource at deploy time (which allows for your API to grow without too much impact to the cloudformation template size), but this means SAM can't find the definition while starting locally.

If there is any low-hanging fruit that could progress this ticket I'll glady help!

### Acknowledgements

### PDK version used

latest/all

_No response_

N/A

cogwirrel:
This adds a new cdk context option `type-safe-api-local` which when set, will use an inline API definition rather than a reference to S3. When synthesized with this context option, the generated cloudformation template is compatible with AWS SAM CLI and can be used with the `sam local start-api` command to run a local development server.

fix #317

### DIFF
--- a/packages/type-safe-api/README.md
+++ b/packages/type-safe-api/README.md
@@ -55,7 +55,7 @@ const api = new TypeSafeApiProject({
   }
 });
 
-// Create a CDK infrastructure project
+// Create a CDK infrastructure project. Can also consider PDKPipelineTsProject as an alternative!
 const infra = new AwsCdkTypeScriptApp({ ... });
 
 // Infrastructure can depend on the generated API infrastructure and runtime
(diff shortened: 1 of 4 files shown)


agdimech (approved):

zetashift:
This is awesome!! Thank you so much :D

PR #373, opened 2023-04-27
chore: upgrade projen to ensure pnpm resolution mode is highest by default

cogwirrel:
Reduces issues when using PNPM, since the default resolution mode is "lowest"

See: https://github.com/projen/projen/issues/2628


.npmrc
```diff
@@ -1,3 +1,4 @@
 # ~~ Generated by projen. To modify, edit .projenrc.ts and run "npx projen".
 
+resolution-mode=highest
 yes=true
```
JeremyJonas, line 4: Is it not sufficient to only have this config set in the root `.npmrc` file? Do we need this in every sub-package as well?

JeremyJonas (commented):
Where is `resolution-mode=highest` being added to all the npmrc in the code? I can't find it in the changes.

.npmrc
cogwirrel: It's set by projen in the latest version - I sent a PR to address it in projen as it affects more than just PDK :) https://github.com/projen/projen/pull/2632

It should be sufficient to just have it in the root - I didn't add any logic like that in my projen change though. I suppose we can remove the `.npmrc` file for all subprojects as part of the nx-monorepo's presynth... but that might be annoying for users if they want to create one on their subproject for whatever reason, so not sure it's worth doing :)

JeremyJonas: Ok sounds good - maybe an area to improve for GA refactor. Since we use monorepo, should only need to maintain root .npmrc for this.

JeremyJonas (approved):

Issue #381, opened 2023-05-01
[BUG] `npm ci` always fails when pdk is installed


tmokmss:
### Describe the bug

Similar to #282, `npm ci` raises an error when used with pdk, but this time I have found no workaround as long as we use npm. Please see the reproduction step.

Because [`npm ci`](https://docs.npmjs.com/cli/v9/commands/npm-ci) is a crucial command for npm users to install dependencies without modifying `package-lock.json`, this issue effectively makes pdk unusable for npm users.

I see PDK assumes to be used with pnpm, but is the above behavior expected? I feel it is unreasonable to shut out npm, which is a default package manager for Node.js.

### Expected Behavior

`npm ci` successes.

### Current Behavior

`npm ci` fails with an error.

### Reproduction Steps

```sh
# run the below commands in an empty directory:
$ npm init -y
$ npm i aws-prototyping-sdk
$ npm i # run this to avoid the error reported on #282
$ npm ci # still get an error below

npm ERR! code EUSAGE
npm ERR!
npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR!
npm ERR! Missing: semver@7.3.4 from lock file
```

### Possible Solution

This bundledDependencies part may be working unexpectedly in npm, as in https://github.com/aws/aws-prototyping-sdk/issues/282#issuecomment-1489757347.

https://github.com/aws/aws-prototyping-sdk/blob/cf57586f2567a571c734606d6f003f6a5051bfca/private/projects/nx-monorepo-project.ts#L22-L27

### PDK version used

0.17.4

### Environment details (OS name and version, etc.)

macOS 12.6.5

JeremyJonas:
Hi @tmokmss, the PDK project (this github repo) itself requires PNPM, but the nx-monorepo project does not require PNPM and should default to projen default packageManager of YARN. 

Can you try using the individual packages rather than bundled `aws-prototyping-sdk` package (eg: `@aws-prototyping-sdk/nx-monorepo`). The bundled package seems to be having issues and we plan to deprecate it in near future.

tmokmss:
Hi @JeremyJonas, you saved my day, thanks! Yeah I also think the bundled package should be deprecated and clearly described in README to avoid confusion.

agdimech:
Closed as part of: https://github.com/aws/aws-prototyping-sdk/pull/384

[...]

Issue #347, opened 2023-04-16
[BUG] jsii-pacmak ERR_FS_FILE_TOO_LARGE when building


jstrunk:
### Describe the bug

When using pnpm to build the package with changes, the build failed on the jsii-pacmak step with `EFF_FS_FILE_TOO_LARGE`. After removing my changes, the error continued. The error went away after I copied my changes onto a clean clone of the repo.

### Expected Behavior

The build should have completed successfully.

### Current Behavior

```
@aws-prototyping-sdk/pipeline: 👾 build » package » package-all » package:js | jsii-pacmak -v --target js --pack-command='pnpm pack'
@aws-prototyping-sdk/pipeline: [jsii-pacmak] [INFO] Found 1 modules to package
@aws-prototyping-sdk/pipeline: [jsii-pacmak] [INFO] Packaging NPM bundles
@aws-prototyping-sdk/pipeline: Error: Command (pnpm pack ) failed with status 1:
@aws-prototyping-sdk/pipeline: #STDERR> 
@aws-prototyping-sdk/pipeline: #STDOUT>  ERR_FS_FILE_TOO_LARGE  File size (2731119967) is greater than 2 GiB
@aws-prototyping-sdk/pipeline: #STDOUT> 
@aws-prototyping-sdk/pipeline: #STDOUT> pnpm [ERR_FS_FILE_TOO_LARGE]: File size (2731119967) is greater than 2 GiB
@aws-prototyping-sdk/pipeline: #STDOUT>     at new NodeError (node:internal/errors:399:5)
@aws-prototyping-sdk/pipeline: #STDOUT>     at tryCreateBuffer (node:fs:435:13)
@aws-prototyping-sdk/pipeline: #STDOUT>     at Object.readFileSync (node:fs:480:14)
@aws-prototyping-sdk/pipeline: #STDOUT>     at packPkg (/Users/<user>/.nvm/versions/node/v18.15.0/lib/node_modules/pnpm/dist/pnpm.cjs:211773:56)
@aws-prototyping-sdk/pipeline: #STDOUT>     at async Object.handler [as pack] (/Users/<user>/.nvm/versions/node/v18.15.0/lib/node_modules/pnpm/dist/pnpm.cjs:211734:7)
@aws-prototyping-sdk/pipeline: #STDOUT>     at async /Users/<user>/.nvm/versions/node/v18.15.0/lib/node_modules/pnpm/dist/pnpm.cjs:221644:21
[...]

PR #384, opened 2023-05-02
fix: resolve issues related to upgrading nx >= 16

### LINKED ISSUE #379: [BUG] Error resolving versions of dependencies in a monorepo
### Describe the bug

What initially started out as projen issue seems to affect `nx-monorepo` as well: https://github.com/projen/projen/issues/2641

When running `npx projen` I'm getting the following error as a result of a recent `projen` upgrade:

```
When running npx projen in my mono repo I'm getting the following error:

foo main• ↑4 ❱ npx projen                                         devenv-shell-env
👾 default | ts-node --project tsconfig.dev.json .projenrc.ts
👾 unable to resolve version for api-typescript-runtime from installed modules
👾 unable to resolve version for api-typescript-infra from installed modules
👾 unable to resolve version for api-typescript-runtime from installed modules
```

@mrgrain said the following:

> It pretty much comes down to this line:
> https://github.com/aws/aws-prototyping-sdk/blob/cf57586f2567a571c734606d6f003f6a5051bfca/packages/nx-monorepo/src/nx-monorepo.ts#L766
> 
> projen core now (correctly) expects that subproject dependencies are installed before resolveDepsAndWritePackageJson is called. Overwriting installDependencies on the subproject has been the go to way for monorepos to do this. But of course there's a reason that method is private.
> 
> I think what this highlights is the missing support for workspace-based monorepos in NodeProject.
> I wonder what the API for this should look like.

### Expected Behavior

The dependencies specified in subprojects being resolved.

### Current Behavior

The packages aren't available because the versions can't be resolved.

### Reproduction Steps

Create a new nx-monorepo

```
mkdir repo && npx projen new --from @aws-prototyping-sdk/nx-monorepo
```

Change the `project` variable in `.projenrc.ts` to this:

```typescript
const project = new NxMonorepoProject({
  defaultReleaseBranch: "main",
  devDeps: ["@aws-prototyping-sdk/nx-monorepo", "@aws-prototyping-sdk/type-safe-api"],
  name: "foo",
  projenrcTs: true,

  deps: [],                /* Runtime dependencies of this module. */
  // description: undefined,  /* The description is just a string that helps people understand the purpose of the package. */
  // packageName: undefined,  /* The "name" in package.json. */
});
```

Then run `npx projen`.

Then change the .projenrc.ts to:

```typescript
import { NxMonorepoProject } from "@aws-prototyping-sdk/nx-monorepo";
import { AwsCdkTypeScriptApp } from "projen/lib/awscdk";
import {
  DocumentationFormat,
  Language,
  ModelLanguage,
  TypeSafeApiProject,
} from "@aws-prototyping-sdk/type-safe-api";

const project = new NxMonorepoProject({
  defaultReleaseBranch: "main",
  devDeps: ["@aws-prototyping-sdk/nx-monorepo", "@aws-prototyping-sdk/type-safe-api"],
  name: "foo",
  projenrcTs: true,

  deps: [],                /* Runtime dependencies of this module. */
  // description: undefined,  /* The description is just a string that helps people understand the purpose of the package. */
  // packageName: undefined,  /* The "name" in package.json. */
});


/** Contains all the smithy model files and codegenerates documentation, clients and API types.
 */
const api = new TypeSafeApiProject({
  name: "api",
  parent: project,
  outdir: "packages/api",
  runtime: {
    languages: [Language.TYPESCRIPT],
  },
  infrastructure: {
    language: Language.TYPESCRIPT,
  },
  documentation: {
    formats: [DocumentationFormat.HTML_REDOC],
  },
  model: {
    language: ModelLanguage.SMITHY,
    options: {
      smithy: {
        serviceName: {
          namespace: "foo.api",
          serviceName: "FooApi",
        },
      },
    },
  },
});

/** Our CDK based project that will deploy our API on AWS Lambda.
 * Also contains all the code for the Lambda handlers.
 */
new AwsCdkTypeScriptApp({
  name: "infra",
  parent: project,
  outdir: "packages/infra",
  defaultReleaseBranch: "main",
  cdkVersion: "2.76.0",
  deps: [
    api.runtime.typescript!.package.packageName,
    api.infrastructure.typescript!.package.packageName,
    "node-fetch",
  ],
  devDeps: [
  ],
  eslint: false,
  gitignore: ["edgeql-js"],
});

project.synth();
```

And run `npx projen` again. It should fail to resolve the dependencies of the `AwsCdkTypeScriptApp`

### Possible Solution

I'm not sure, but this seems like it might require an API change for `NodeProject`s.

See https://github.com/projen/projen/issues/2641 and the comments by mrgrain that highlight the way `projen` dealt with version resolution before and after the latest version release.

### PDK version used

0.17.4

Typescript

MacOS ARM64 13.3.1

agdimech:
Resolve issues related to upgrading nx >= 16

- Task execution order is now correct due to targetDefaults being set correctly
- Updates to deprecated @nrwl packages
- Pin nx as a peer dep to >= 16
- Package versions are now resolved correctly due to recent change in Projen.
- Removed uber package and refactored samples to take dependencies on individual packages

BREAKING CHANGE: Deprecate uber aws-prototyping-sdk package

Fixes #379, #381, #347


nx.json
```diff
@@ -40,34 +40,16 @@
...
-        "projects": "dependencies"
-      }
-    ],
     "install": [
       {
         "target": "install",
         "projects": "dependencies"
       }
     ]
   },
```
JeremyJonas: Should this be moved to `targetDefaults` as well?

nx.json
```diff
@@ -40,34 +40,16 @@
       ]
     }
   },
+  "affected": {
+    "defaultBase": "mainline"
+  },
   "targetDependencies": {
```
JeremyJonas: This should be removed right? It has been deprecated?

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -451,15 +450,11 @@ export class NxMonorepoProject extends TypeScriptProject {
...
-              target: "build",
-              projects: "dependencies",
-            },
-          ],
-          ...(this.nxConfig?.targetDependencies || {}),
+        targetDefaults: {
+          build: {
+            dependsOn: ["^build"],
+          },
+          ...(this.nxConfig?.targetDefaults || {}),
```
JeremyJonas: FYI: `|| {}` is redundant for object spread, only necessary for array spread.

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -776,8 +773,13 @@ export class NxMonorepoProject extends TypeScriptProject {
...
     ) {
-      // @ts-ignore - `installDependencies` is private
-      this.package.installDependencies();
+      try {
+        // @ts-ignore - `installDependencies` is private
+        this.package.installDependencies();
+      } finally {
+        // @ts-ignore - `resolveDepsAndWritePackageJson` is private
+        this.package.resolveDepsAndWritePackageJson();
+      }
```
JeremyJonas, line 782: Missing a `catch` here?

This logic will always run `installDependencies` and `resolveDepsAndWritePackageJson`, even if `installDependencies` fails, but the closure will still fail afterwards if `installDependencies` fails.

I think you want an empty catch here.

JeremyJonas (approved):
Awesome, thank you. Approved with comments, which if are as expected can merge.

nx.json
agdimech: I don't believe it has as per: https://nx.dev/reference/nx-json

nx.json
agdimech: Ahh yep - good catch

packages/nx-monorepo/src/nx-monorepo.ts
agdimech: Nice.

packages/nx-monorepo/src/nx-monorepo.ts
agdimech: It's actually correct as it - I want to throw but need to make sure the package.json is written out so it doesn't end up in a corrupt state.

PR #394, opened 2023-05-04
feat(type-safe-api): react query hooks library

cogwirrel:
Adds support for generation of libraries, which are not fully fledged runtimes. The library this change introduces is a typescript react-query hooks library, which generates a hook per API operation for use in a react UI.

Note also that `useInfiniteQuery` hooks are generated for operations which use the `x-paginated` vendor extension to indicate the input/output tokens for pagination. With Smithy, at present adding vendor extensions requires using `jsonAdd` in `smithy-build.json`, however we can improve the user experience with a trait once https://github.com/awslabs/smithy/pull/1609 is merged and released.


packages/type-safe-api/README.md
```diff
@@ -1035,6 +1041,146 @@ public class SayHelloHandler extends SayHello {
...
+  library: {
+    libraries: [Library.TYPESCRIPT_REACT_QUERY_HOOKS],
+  },
+  ...
+});
+```
+
+##### Usage in a React Website
+
+First, make sure you create an instance of the API client (making sure to set the base URL and fetch instance). For example:
```
agdimech, line 1063: NIT: Probably should call out that a dependency would be needed on the lib package.

agdimech (approved):

agdimech:
Great Job. Looking forward to testing this out once released :)

PR #396, opened 2023-05-05
fix: bump dependencies to use TS 5.X

agdimech:
I feel the need, the need for speed!

## What's new?

- Bumps deps to use JSII 5.x and Ts 5.x which means Jest ^29 is now possible which means faster test executions :)
- Projen PR https://github.com/projen/projen/pull/2653 has been merged so this PR will yield HUGE performance improvements ~60% improvement in build times and increased CPU utilization (10 min local build without caching).
- Prefer local workspace packages has been enabled in npmrc.
- Removed patch-tsviz from cdk-graph-plugin-diagram now that TS5.X is being used.
- Had to increase the failureThreshold in test-utils.ts for png comparison to 1.5%. @JeremyJonas please confirm this is OK.

## Actions

- **@jsii-ignore is needed in a few places due to unsupported interface features which were previously silently failing for non-js languages. We need to refactor this as this won't work for py/java as per https://aws.github.io/jsii/compiler-and-rosetta-maintenance/#how-difficult-is-it-to-migrate-from-1x-to-50x.** -- #399


packages/cdk-graph-plugin-diagram/src/internal/graphviz/theme/index.ts
```diff
@@ -367,6 +367,7 @@ const GRAPH_ATTRIBUTES: Dot.GraphAttributesObject = {
   labelloc: "tc",
   nodesep: 0.6,
   pad: "%2,%1",
+  // @ts-ignore
```
agdimech, line 370: compiler was complaining that TB wasn't an option for some reason even though it is the default.

packages/cdk-graph-plugin-diagram/test/graphviz/test-utils.ts
```diff
@@ -53,8 +53,8 @@ export async function expectToMatchImageSnapshot(
       // Prevent rendering variants between environments (CI, MacOS, Ubuntu, etc)
       threshold: pixelThreshold || 0.05, // default is 0.01
     },
-    // Allow a 1% difference in image for testing
-    failureThreshold: failureThreshold || 0.01,
+    // Allow a 1.5% difference in image for testing
+    failureThreshold: failureThreshold || 0.015,
```
agdimech, line 57: Had to bump this as the difference was like 1.1% :( @JeremyJonas - can you confirm this is OK please?

private/pdk-project.ts
```diff
@@ -111,8 +102,13 @@ export abstract class PDKProject extends JsiiProject {
         }) ||
         undefined,
       gitignore: [...(options.gitignore || []), "LICENSE_THIRD_PARTY"],
+      disableTsconfigDev: false,
+      disableTsconfig: true,
     });
 
+    this.preCompileTask.prependExec("rm -f tsconfig.json");
```
agdimech, line 109: jsii creates a tsconfig file but doesn't clean up after itself 😠

JeremyJonas (approved):

PR #400, opened 2023-05-10
feat(nx-monorepo): nx composition and caching improvements

JeremyJonas:
- fix(nx-monorepo): resolve postinstall for non-node project
- refactor(nx-monorepo): use public project.subprojects
- chore: upgrade project to ^0.71.48
- feat(nx-monorepo): infer nx input/output by project type
  - enabled only for pdk repo, but can be enabled by consumers as well as is still experimental
- refactor(nx-monorepo): make nx configs composible
- feat(nx-monorepo): make NX_NON_DEFAULT_HASHER opt-in
  - Now we use the default nx hasher by default


packages/nx-monorepo/src/components/nx-project/index.ts
```diff
@@ -0,0 +1,277 @@
...
+  "preinstall",
+  "install",
+  "postinstall",
+  "preinstall:ci",
+  "install:ci",
+  "postinstall:ci",
+];
+
+/**
+ * Component which manged the project specific NX Config and is added to all NXMonorepo subprojects.
```
JeremyJonas: *manages

packages/nx-monorepo/src/components/nx-project/index.ts
```diff
@@ -0,0 +1,277 @@
...
+   * Implicit dependencies
+   *
+   * @see https://nx.dev/reference/project-configuration#implicitdependencies
+   */
+  public implicitDependencies: string[] = [];
+  /**
+   * Explicit list of scripts for Nx to include.
+   * @see https://nx.dev/reference/project-configuration#ignoring-package.json-scripts
+   */
+  public includedScripts: string[] = [];
```
JeremyJonas, line 84: Need to implement this in synth below - forgot to do this

packages/nx-monorepo/src/components/nx-project/index.ts
```diff
@@ -0,0 +1,277 @@
...
+    // Make sure we only ever have 1 instance of NxProject component per project
+    if (NxProject.of(project))
+      throw new Error(
+        `Project ${project.name} already has associated NxProject component.`
+      );
+
+    super(project);
+
+    this.file =
+      (project.tryFindObjectFile("project.json") as JsonFile) ||
```
JeremyJonas: If "project.json" already exists, then the below `obj` will not get synthesized - need to fix this

packages/nx-monorepo/src/components/nx-project/targets.ts
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Specifies how `dependsOn` for the target is derived.
+   * - If `false`, no `dependsOn` will be added.
+   * - If `true`, the default `^build` will be used.
+   * - If explicit list of strings, the value will be used.
+   */
+  readonly dependsOn?: boolean | string[];
+}
+
+export function inferBuildTarget(
```
JeremyJonas, line 131: docs

packages/nx-monorepo/src/components/nx-workspace.ts
```diff
@@ -0,0 +1,243 @@
...
+  DependencyType,
+  IgnoreFile,
+  JsonFile,
+  Project,
+} from "projen";
+import { Obj } from "projen/lib/util";
+import { Nx } from "../nx-types";
+import { asUndefinedIfEmpty, deepMerge } from "../utils/common";
+
+export const ALWAYS_IGNORE: string[] = [".tmp", ".env", ".pytest_cache"];
```
JeremyJonas: docs

packages/nx-monorepo/src/components/nx-workspace.ts
```diff
@@ -0,0 +1,243 @@
...
+  "{projectRoot}/build",
+  "{projectRoot}/coverage",
+  "{projectRoot}/test-reports",
+  "{projectRoot}/target",
+  "{projectRoot}/LICENSE_THIRD_PARTY",
+  "{projectRoot}/.jsii",
+];
+
+/**
+ * Component which manged the workspace specific NX Config for the root monorepo.
```
JeremyJonas: *manages

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -89,10 +89,10 @@ export interface MonorepoUpgradeDepsOptions {
...
  */
 export interface NxMonorepoProjectOptions extends TypeScriptProjectOptions {
-  /**
-   * Configuration for NX.
-   */
-  readonly nxConfig?: Nx.WorkspaceConfig;
+  // /**
+  //  * Configuration for NX.
+  //  */
+  // readonly nxConfig?: Nx.WorkspaceConfig;
```
JeremyJonas: Remove this

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -667,9 +549,30 @@ export class NxMonorepoProject extends TypeScriptProject {
     (this.tasks.tryFind("prepare") || this.addTask("prepare")).spawn(linkTask);
   }
 
+  /**
+   * Ensures that all non-root projects have NxProjectConfig applied.
```
JeremyJonas: *NxProject

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -667,9 +549,30 @@ export class NxMonorepoProject extends TypeScriptProject {
...
+      });
+    }
+
+    this.subprojects.forEach(_ensure);
+  }
+
   preSynthesize(): void {
     NodePackageUtils.removeProjenScript(this);
 
+    // Calling being "super" to ensure proper preSynth of NxProject component and its nested components
```
JeremyJonas: *before

packages/nx-monorepo/src/nx-types.ts
```diff
@@ -1,37 +1,60 @@
...
 SPDX-License-Identifier: Apache-2.0 */
+
+import { Obj } from "projen/lib/util";
+
 export namespace Nx {
   /**
    * Configuration for nx targetDependencies.
    */
-  export type TargetDependencies = { [target: string]: TargetDependency[] };
+  export type ITargetDependencies = { [target: string]: ITargetDependency[] };
```
JeremyJonas: From "I" prefix, just a type

packages/nx-monorepo/src/nx-types.ts
```diff
@@ -1,37 +1,60 @@
...
 
   /**
    * Configuration for project specific targets.
    */
-  export type ProjectTargets = { [target: string]: ProjectTarget };
+  export type IProjectTargets = { [target: string]: IProjectTarget };
```
JeremyJonas: Remove "I" prefix, just a type

packages/cdk-graph/src/config/index.ts
```diff
@@ -13,7 +13,7 @@ export const CDK_OUTDIR_VAR = "<cdk.out>";
 export const DEFAULT_OUTDIR = `${CDK_OUTDIR_VAR}/cdkgraph`;
 
 /** CdkGraph configuration definition */
-export interface CdkGraphConfig {
+export type CdkGraphConfig = {
```
agdimech, line 16: Do types means index types will work cross language?

packages/nx-monorepo/src/components/nx-project/index.ts
```diff
@@ -0,0 +1,304 @@
...
+          // exclude node lifecycle tasks for node based projects
+          return false;
+        }
+        return true;
+      })
+      .forEach((task) => {
+        const _target = this.targets[task.name] || {};
+        _target.executor = _target.executor || "nx:run-commands";
+        _target.options = {
+          command: NodePackageUtils.command.downloadExec(
```
agdimech: why do we need download exec here? I would assume projen is already installed so we should just need to call exec?

packages/nx-monorepo/src/utils/common.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Recursively merge objects together into a new object.
+ *
+ * By default, arrays are not merged
+ * but overwritten just like scalars. Use the `append=true` flag to append array values.
+ *
+ * If an object is merged into a non-object, the non-object is lost.
+ *
+ * `undefined`s will cause a value to be deleted if destructive is enabled.
+ */
+export function deepMerge(
```
agdimech, line 60: Is the different to the deepMerge in projen/lib?

packages/nx-monorepo/src/utils/node.ts
```diff
@@ -19,6 +24,14 @@ export namespace NodePackageUtils {
     return cmd;
   }
 
+  /** Indicates if project is a node based project */
+  export function isNodeProject(project: Project): boolean {
+    if (project instanceof NodeProject) {
```
agdimech: Can be simplified to:

`return project instanceof NodeProject || tryFindNodePackage(project, false) != null;`

private/pdk-project.ts
```diff
@@ -182,86 +183,50 @@ export abstract class PDKProject extends JsiiProject {
...
-   * @param value Value to override
-   * @param {boolean} [append=false] Indicates if array values are appended to, rather than overwritten.
-   */
-  public nxOverride(path: string, value: any, append?: boolean): void {
-    const nx = cloneDeep(this.nx);
-    overrideField(nx, path, value, append);
-
-    this.nx = nx;
-  }
+  // /**
```
agdimech, line 194: Any reason this just can't be deleted?

private/pdk-project.ts
```diff
@@ -345,16 +310,12 @@ class PDKRelease extends Release {
...
-}
 
-/**
- * Utility to deeply clone a value
- * @param value Value to clone
- * @returns Cloned value
- */
-export function cloneDeep(value: any): any {
-  return JSON.parse(JSON.stringify(value));
+    NxProject.of(project)?.addBuildTargetFiles(
```
agdimech, line 313: Can't this just be:

`NxProject.of(project)?.addBuildTargetFiles(
      [],
      ["{projectRoot}/LICENSE_THIRD_PARTY"],
      true
    );`

private/projects/pdk-monorepo-project.ts
```diff
@@ -90,33 +54,26 @@ export class PDKMonorepoProject extends NxMonorepoProject {
...
-        },
-      },
       workspaceConfig: {
         disableNoHoistBundled: true,
         linkLocalWorkspaceBins: true,
       },
     });
 
+    // Turn on automatic target inference for this repo
+    this.nx._autoInferProjectTargets = true;
```
agdimech: Is there a nicer way to set this?

private/projects/pdk-monorepo-project.ts
```diff
@@ -90,33 +54,26 @@ export class PDKMonorepoProject extends NxMonorepoProject {
...
+    // This is OK to be stored given its read only and the repository is public
+    this.nx.useNxCloud(
+      "OWJmZDJmZmEtNzk5MC00OGJkLTg3YjUtNmNkZDk1MmYxZDZkfHJlYWQ="
+    );
+    this.nx.cacheableOperations.push("generate");
```
agdimech, line 69: Can we also set the cacheDirectory to `.nx/cache`?

packages/nx-monorepo/src/nx-monorepo.ts
```diff
@@ -363,70 +316,15 @@ export class NxMonorepoProject extends TypeScriptProject {
...
-   * Ensure NXProject is added when adding subprojects to the monorepo.
-   *
-   * @internal
-   * @param subproject project to add.
-   */
-  _addSubProject(subproject: Project) {
-    !NxProject.of(subproject) &&
-      subproject._addComponent(new NxProject(subproject));
-    super._addSubProject(subproject);
+    this.nx = NxWorkspace.of(this) || new NxWorkspace(this);
```
agdimech, line 319: Can we also have an ensure method in NxWorkspace?

agdimech (commented):
Really great improvements mate!

packages/cdk-graph/src/config/index.ts
JeremyJonas: Using `type` gives the following JSII, which is `map` - so looks like should work right?

```
        {
          "abstract": true,
          "docs": {
            "see": "https://nx.dev/reference/project-configuration",
            "stability": "stable",
            "summary": "Targets configuration."
          },
          "immutable": true,
          "locationInModule": {
            "filename": "src/nx-types.ts",
            "line": 246
          },
          "name": "targets",
          "optional": true,
          "type": {
            "collection": {
              "elementtype": {
                "fqn": "@aws-prototyping-sdk/nx-monorepo.Nx.IProjectTarget"
              },
              "kind": "map"
            }
          }
        }
      ],
```

packages/nx-monorepo/src/components/nx-project/index.ts
JeremyJonas: I copied it from the code you wrote before 😊 ... assuming was to do with non-node based package/repo support? Since before was only used for non-node projects.

packages/nx-monorepo/src/utils/common.ts
JeremyJonas: Yes, it adds functionality to "append to array"... I will update description to make this clear

private/pdk-project.ts
JeremyJonas: Oh it should have been, not sure how missed that

packages/nx-monorepo/src/nx-monorepo.ts
JeremyJonas: NxWorkspace should only be set once by the root, and is not required. While do to construct/synth order of operations, the NxProject might not have been instantiated when it is needed - so `ensure` makes sense. I don't think we want the same loose functionality around NxWorkspace.

But if have a use case for it then yes is easy to add

private/projects/pdk-monorepo-project.ts
JeremyJonas: Ah yes, forgot about that

private/projects/pdk-monorepo-project.ts
JeremyJonas: What you have in mind? it is experimental and I expect we will enable it by default once we have more robust coverage and feedback - at which time was planning to change, which I why I am ok with an ugly implementation which stands out a bit.

private/pdk-project.ts
JeremyJonas: That will generate inputs of `["!{projectRoot}/LICENSE_THIRD_PARTY", "!{projectRoot}/LICENSE_THIRD_PARTY/**/*"]`, because we don't know if it is a file or directory. Since we know this is just a file it is best to just be explicit.

NX input seems to require the `/**/*` for directories.

packages/nx-monorepo/src/components/nx-project/index.ts
JeremyJonas: I have removed the `dlx` stuff - using `NodePackageUtils.command.projen(packageManager, task.name)` now

private/pdk-project.ts
agdimech: How about:

`NxProject.of(project)?.addBuildTargetFiles( [], [File.of("{projectRoot}/LICENSE_THIRD_PARTY")], true );`

private/projects/pdk-monorepo-project.ts
agdimech: `this.nx.autoInferProjectTargets()`

packages/nx-monorepo/src/nx-monorepo.ts
agdimech: Makes sense, please ignore.

agdimech (approved):
Great work!

PR #401, opened 2023-05-11
feat(type-safe-api): add a DefaultApiClientProvider which handles sha…

agdimech:
The @tanstack/react-query package does not play nice when writing provider/use code across multiple packages. To workaround this we use a singleton context by default that is shared between provider (customer code) and consumer (hooks code). This ensures the client is always available in the correct context whilst simplifying the interface that the customer has to code against.

To get started all they need to do is:

```
// Somewhere in a parent component
import { DefaultApiClientProvider as DefaultApiClientProvider_ } from "myapi-typescript-react-query-hooks";

<DefaultApiClientProvider apiClient={client}>
      {children}
</DefaultApiClientProvider>
```

This removes the need for customer's to take an additional dependency on @tanstack-react-query :)

### DIFF
--- a/packages/type-safe-api/README.md
+++ b/packages/type-safe-api/README.md
@@ -74,20 +74,20 @@ monorepo.synth();
 In your CDK application (ie within the `infra` project we created), consume the `Api` construct, vended from the generated typescript infrastructure package.
 
 ```ts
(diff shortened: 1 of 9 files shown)


JeremyJonas (approved):
Awesome!

cogwirrel:
Nice one, thank you!

PR #402, opened 2023-05-11
feat(type-safe-api): namespace query client context and update docs

cogwirrel:
We namespace the query client context (prefixing with the API client classname), so that we don't get clashes if the user splits operations into different APIs with tags.
    
Additionally add some more documentation to the readmes.


packages/type-safe-api/README.md
```diff
@@ -1245,6 +1255,20 @@ new TypeSafeApiProject({
...
 
+##### Custom QueryClient
+
+If you wish to customise the react-query `QueryClient`, pass a custom instance to the client provider, eg:
+
+```tsx
+const queryClient = new QueryClient({ ... });
+
+return (
+  <MyApiClientProvider apiClient={api} client={queryClient}>
```
agdimech: For me this is DefaultApiClientProvider even thought I have its name set as MyApi. But S that a bug?

cogwirrel: Ah so by default it'll be called `DefaultApiClientProvider` (just like the client is called `DefaultApi` by default), but it's named based on operation tags. Eg if you tagged your operations `PetStore` it'd be called `PetStoreApiClientProvider` :)

Wanted to make it clearer in the docs that it might not always be called `Default`, but maybe it adds confusion! Can rename to `DefaultApi` in docs if you prefer? :)

cogwirrel: Renamed to `DefaultApi` with a note about it being different if you've tagged your operations :)

agdimech (approved):

PR #410, opened 2023-05-16
feat(nx-monorepo): impl independent versioning and release

JeremyJonas:
- implement independent package versioning and release (NxRelease, NxReleaseProject)
- will automatically start all packages from 0.5.0 (which will bump to 0.6.0), since we are around 0.4.3 at time of writing, and will be diffed against 0.4.3 for change logs.

Tags and Releases have been published to https://github.com/JeremyJonas/aws-prototyping-sdk/releases from local `pnpm release` execution.

> I have not tested the publish tasks, but they are basically unchanged so expected to work

---
## ATTENTION: merge process for this release
- [x] Disable release.yaml workflow
- [x] Create last-release tag on the latest releases commit
- [x] Create a {package-name}-{version} tag based on latest v0.x released version for each package (eg: @aws-prototyping-sdk/nx-monorepo-0.18.2) - version is without v prefix
- [x] Merge this PR
- [x] Re-enable release.yaml workflow, and manually trigger it if necessary.

> DO NOT MERGE ANY OTHER CHANGE ONCE THIS PROCESS STARTS


.github/workflows/release.yml
```diff
@@ -1,13 +1,11 @@
 name: release
+# Ensure each release is completed before next is performed
+concurrency: release_mainline
```
agdimech, line 3: Nice improvement.

.github/workflows/release.yml
```diff
@@ -49,122 +77,16 @@ jobs:
...
-        env:
-          TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
-          TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
   release_docs:
     needs: release_github
     runs-on: ubuntu-latest
     permissions:
       contents: write
     if: needs.release.outputs.latest_commit == github.sha
+    env:
```
agdimech, line 76: Why do we need these given we are no longer zipping up and gh releasing?

packages/nx-monorepo/src/components/nx-workspace.ts
```diff
@@ -228,13 +245,35 @@ export class NxWorkspace extends Component {
     );
   }
 
-  /** @internal */
-  protected _recursivelyApplyToProject(
+  /**
+   * Recursively apply a function a project and subprojects.
```
agdimech: a function to*

packages/nx-monorepo/src/components/nx-workspace.ts
```diff
@@ -228,13 +245,35 @@ export class NxWorkspace extends Component {
...
+   * @param predicate Predicate to match projects to apply function to
+   */
+  public recursivelyApplyToProject(
     project: Project,
-    fn: (project: Project) => void
+    fn: ProjectCallback,
+    predicate?: ProjectPredicateCallback
+  ): void {
+    (!predicate || predicate(project)) && fn(project);
+    project.subprojects.forEach((p) =>
```
agdimech: could this be: `recursivelyApplyToProjects(project.subprojects, fn, predicate);`

private/projects/pdk-monorepo-project.ts
```diff
@@ -143,6 +144,10 @@ export class PDKMonorepoProject extends NxMonorepoProject {
     this.addTask("dev:workspace:unlink", {
       exec: "./scripts/dev/link-workspace.js unlink",
     });
+
+    const nxRelease = NxRelease.of(this) || new NxRelease(this);
+    // Pin initial versioned release to last non-versioned release commit
+    nxRelease.initialReleaseStart = "d57fee1447943fc848980d9a6036710d3236f387";
```
agdimech: We should bump this to latest, given a few commits have been merged recently.

packages/nx-monorepo/scripts/release-version-sync.ts
```diff
@@ -0,0 +1,62 @@
...
+/*! Copyright [Amazon.com](http://amazon.com/), Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0 */
+import * as path from "node:path";
+import * as fs from "fs-extra";
+import { SemVer } from "semver";
+import { createProjectGraphAsync } from "@nrwl/devkit";
+
+const DEPS_PROPS = ["dependencies", "devDependencies", "peerDependencies", "optionalDependencies"]
+
+async function releaseVersionSync() {
```
agdimech, line 18: Add some jsdocs here please.

agdimech (commented):
Great set of improvements Jeremy.

One thing that wasn't clear to me was how we handle promotion from 0.X -> 1.0.0? I understand that for 1.0.0 or above that will just be done automatically via a BREAKING-CHANGE guidance in the commit msg but what happens of 0.X?

Also - do we think we should also bring up the commitizen/commit-lint checks into nx-monorepo as part of this PR or do that as a fast follow?

.github/workflows/release.yml
```diff
@@ -35,9 +54,18 @@ jobs:
...
         run: git diff --ignore-space-at-eol --exit-code
+      - name: Tag last-release
+        shell: bash
+        run: |
+          git tag -f last-release
+          git push origin last-release --force
+      - name: Print release tags
+        run: git tag --list --contains `git rev-parse HEAD`
+      - name: Push release tags
+        run: git push origin --tags
```
JeremyJonas: Need to remove this - we do not want to push all local tags since could include "minVersion" markers.

.github/workflows/release.yml
JeremyJonas: GH release above is just for the "releases", this is for publishing the "docs" to github pages. It would be great to also get that functionality into the DocsProject, but due to being a subdir and different branch is non-trivial.

.github/workflows/release.yml
```diff
@@ -180,11 +102,6 @@ jobs:
         run: |-
           git config user.name "AWS PDK Automation"
           git config user.email "aws-pdk+automation@amazon.com"
-      - name: Upload docs to Github
-        run: zip -r docs.zip dist/public/docs/dist/docs/* && gh release upload $(cat dist/packages/nx-monorepo/dist/releasetag.txt) -R $GITHUB_REPOSITORY docs.zip && rm docs.zip
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GITHUB_REPOSITORY: ${{ github.repository }}
       - name: Prepare Commit
```
JeremyJonas, line 94: Need to double check this, should it be removed?

agdimech: This was previously storing the whole docs as a zip so users could view docs at a 'point in time'. I don't think this makes sense any more given we no longer release as a whole. The docs are also now contained within the package so it should be fine to remove.

packages/nx-monorepo/scripts/release-version-post.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const projectRoot = path.join(workspaceRoot, await getProjectRoot(projectName));
+
+    // patch the package.json version
+    const pgkJsonPath = path.join(projectRoot, "package.json");
+    const pkgJson = await fs.readJson(pgkJsonPath);
+    pkgJson.version = version;
+    await fs.writeJson(pgkJsonPath, pkgJson, { spaces: 2 });
+
+    // path the changelog
```
agdimech: patch*

agdimech (commented):
Fantastic work!

Should we add a PR Lint task/commit hook to ensure that breaking change commits only affect package + root?

agdimech (approved):

JeremyJonas:
- [x] Disable release.yaml workflow
- [x] Create last-release tag on the latest releases commit
- [x] Create a {package-name}-{version} tag based on latest v0.x released version for each package (eg: @aws-prototyping-sdk/nx-monorepo-0.18.2) - version is without v prefix
- [ ] Merge this PR
- [ ] Re-enable release.yaml workflow, and manually trigger it if necessary.

PR #412, opened 2023-05-17
feat(type-safe-api): add mock integration for returning hardcoded responses

cogwirrel:
This adds `Integrations.mock` for hardcoding a response for API operations.

The response will automatically include CORS headers if CORS was enabled on the API. The response provided is mapped to all request content types for the operation (this is transparent to the user given the content types are added to the `OperationLookup` in all languages).

Note that this isn't type safe to the operation's response types (response body is just a string), however this provides a foundation for generating type safe mocks as part of the generated cdk infrastructure package.

eg we could generate something like:

```ts
class MockIntegrations {
  public static sayHello(response: SayHelloResponseContent) {
    return Integrations.mock({
      statusCode: 200,
      body: JSON.stringify(SayHelloResponseContentToJSON(response)),
    });
  }
}
```

This change also refactors individual integrations into separate files.


packages/type-safe-api/test/resources/specs/single.yaml
```diff
@@ -98,6 +98,26 @@ paths:
             application/x-mpegurl:
               schema:
                 type: string
+  /multiple-content-types:
```
agdimech, line 101: fancy

agdimech (approved):

PR #421, opened 2023-05-18
feat(type-safe-api): generate type safe mock integrations in infrastructure packages

cogwirrel:
The generated infrastructure packages now include a `MockIntegrations` class which exposes methods for returning a mock integration for every type of response returned by an operation.

eg:

```ts
new Api(this, "Api", {
  defaultAuthorizer: Authorizers.iam(),
  integrations: {
    sayHello: {
      integration: MockIntegrations.sayHello200({
        message: "mock in typescript!",
      }),
    },
  },
});
```

```java
new Api(s, "Api", ApiProps.builder()
        .integrations(OperationConfig.<TypeSafeApiIntegration>builder()
                .sayHello(TypeSafeApiIntegration.builder()
                        .integration(MockIntegrations.sayHello200(SayHelloResponseContent.builder()
                                .message("mock in java!")
                                .build()))
                        .build())
                .build())
        .build());
```

```python
Api(self, 'Api',
    default_authorizer=Authorizers.iam(),
    integrations=OperationConfig(
        say_hello=TypeSafeApiIntegration(
            integration=MockIntegrations.say_hello_200({
                "message": "mock in python!"
            })
        ),
    ),
)
```


packages/type-safe-api/scripts/generators/java-cdk-infrastructure/templates/mockIntegrations.mustache
```diff
@@ -0,0 +1,45 @@
...
+import software.aws.awsprototypingsdk.typesafeapi.MockIntegration;
+import software.aws.awsprototypingsdk.typesafeapi.MockIntegrationResponse;
+
+/**
+ * Type-safe mock integrations for API operations
+ */
+public class MockIntegrations {
+    static {
+        // Instantiate json instance so that any .toJson() methods can be used
+        new JSON();
```
agdimech, line 15: This is odd. Why do we create the instance without doing anything with it? Is there some janky singleton behaviour going on under the covers or something?

agdimech (approved):

packages/type-safe-api/scripts/generators/java-cdk-infrastructure/templates/mockIntegrations.mustache
cogwirrel: Yeah exactly that - it's super janky haha!

The `.toJson` method in a model calls the static `JSON.getGson()` method ([see here](https://github.com/OpenAPITools/openapi-generator/blob/7881482161d22c138792cb9333daff216d6d68db/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/pojo.mustache#L657)).

But the `gson` instance that `JSON.getGson()` returns is only initialised in a non-static initialiser block - it's only the act of creating some instance of `JSON` somewhere that initialises the `gson` instance:

https://github.com/OpenAPITools/openapi-generator/blob/7881482161d22c138792cb9333daff216d6d68db/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache#L112-L144

agdimech: That is so dumb - they should change that to a static initializer so it gets called the first time the class is loaded by the classloader...

cogwirrel: Yeah! No idea why it's not static... It caught me out when I built the lambda handler wrappers - things only failed at runtime which was super fun to debug 😆

PR #430, opened 2023-05-23
fix(nx-monorepo): release workflow

JeremyJonas:
- skip tests during release
- try to fix git tag permissions

---
- https://github.com/aws/aws-prototyping-sdk/actions/runs/5052996883/jobs/9066401128
- https://github.com/orgs/community/discussions/35410

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,7 @@ jobs:
       NX_BRANCH: mainline
       NX_RUN_GROUP: ${{ github.run_id }}
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+      NX_CLOUD_NO_TIMEOUTS: true
       # GITHUB
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       GITHUB_REPOSITORY: ${{ github.repository }}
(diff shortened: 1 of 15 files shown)


.github/workflows/release.yml
```diff
@@ -50,18 +52,9 @@ jobs:
...
         run: pnpm release
-      - name: Check for mutations
-        run: git diff --ignore-space-at-eol --exit-code
-      - name: Tag last-release
-        shell: bash
-        run: |
-          git tag -f last-release
-          git push origin last-release --force
-      - name: Print release tags
-        run: git tag --list --contains `git rev-parse HEAD`
```
JeremyJonas, line 64: - Removed "check for mutations" because we still want to push docs if nothing was releases.
- The "Tag last-release" is already handled in the "release" task, so removing it here as was duplicate
- Moved "Print release tags" to the release task - just output for all the release tags, not important

.github/workflows/release.yml
```diff
@@ -39,6 +40,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
```
JeremyJonas, line 43: As per https://github.com/orgs/community/discussions/35410 solution

agdimech (approved):

PR #434, opened 2023-05-25
fix(nx-monorepo): resolve dependencies in node subprojects

BradenM:
This is a follow-up to #379 and projen/projen#2641

The current logic in-place essentially skips resolution of sub-project dependencies entirely (as described by @mrgrain in https://github.com/projen/projen/issues/2641#issuecomment-1529093263).

As a result of this, dependencies added to child projects will always be '\*' unless manually pinned or edited with a scope.
Neither `projen upgrade` nor `projen upgrade-deps` will update this for you as of now either (at least from '*' that is), so it requires manual intervention.

In the aim of resolving this problem, this PR modifies the synth process to ensure subprojects will be resolved only after the workspace root has had a chance to install any changes.
This is based on @mrgrain 's mockup seen here: https://github.com/aws/aws-prototyping-sdk/issues/379#issuecomment-1529102385


agdimech:
Good catch! Awesome PR :)

cogwirrel (approved):
Nice work! Thanks for the fix! 😁

PR #435, opened 2023-05-26
fix: symlinks cannot be published to npm

agdimech:
Previously, the codebase was uploaded as an artifact and re-downloaded which means any hardlinks would have been dereferenced as part of this process. npm publish would have worked as all files would be non-linked and as such is only a problem for packages with bundledDeps.

To remedy this, we simply change the link-bundled-transitive deps script to use a copy with dereference to ensure all node_module packages are actuals copied files.

In addition, the PDK init was missing an install of Java and Python runtimes.

### DIFF
--- a/.github/actions/pdk-init/action.yml
+++ b/.github/actions/pdk-init/action.yml
@@ -7,6 +7,13 @@ runs:
       uses: actions/setup-node@v3
       with:
         node-version: 18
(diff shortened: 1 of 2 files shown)


cogwirrel (approved):

JeremyJonas (approved):
Nice!!!

PR #444, opened 2023-05-30
feat(type-safe-api): support for api keys

cogwirrel:
Adds support for api keys via `apiKeyOptions`. API Keys can be required when the API key source is 'HEADER'. Keys can be required by default and/or required/not-required for specific methods.

Note that this change simplifies the way that default authorizers are applied by applying authorizers only at the method level.

I'll add docs for this in my separate docs PR.


agdimech (approved):
lgtm

PR #451, opened 2023-06-01
feat(type-safe-api): generate mock data for use in mock integrations

cogwirrel:
This PR allows users to call the appropriate mock integration method without any arguments to use generated mock data in the response for their mock integration, to save even more time when mocking an API. Note that only non-primitive return types are mocked (ie json structures), other response types must be supplied by the user.

Also fix some of the incorrectly transpiled java examples in the docs.


packages/type-safe-api/docs/developer_guides/type-safe-api/lambda_handlers.md
```diff
@@ -440,62 +440,34 @@ When you use a handler router, you must specify the same lambda function for eve
...
-        super(scope, id, {
-          defaultAuthorizer: Authorizers.iam(),
-          // Use the same integration for every operation.
-          integrations: Operations.all({
-            integration: Integrations.lambda(new NodejsFunction(scope, "router")),
-          }),
-        });
-      }
-    }
+    new Api(this, "Api", {
```
agdimech, line 443: Should we preserve the imports so users know where it comes from?

agdimech (approved):

packages/type-safe-api/docs/developer_guides/type-safe-api/lambda_handlers.md
cogwirrel: The imports were incorrect in the transpiled code - I'll revisit in another PR! :)

PR #471, opened 2023-06-16
feat(type-safe-api): manage settings gradle file to make service name changes easier

cogwirrel:
If you change the name of your api after initial synthesis, you'd previously need to manually edit `settings.gradle` to match your new service name, so that the openapi spec is generated in the directory expected by the `parse-openapi-spec` command. This change moves `settings.gradle` from a sample file to a projen managed file so it's updated automatically.

Additionally, update the default smithy model to include some better errors, more concise operation syntax, and an example directory structure.


JeremyJonas (approved):
🚀 🙇

PR #475, opened 2023-06-20
docs: reviewed type-safe-api topic

swap-aws:
Fixes #


packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
```diff
@@ -1,14 +1,19 @@
 # Type Safe API
 
+_question: What does this refer to?_
```
cogwirrel: Does `_question` render something differently in the docs website? :)

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
```diff
@@ -1,14 +1,19 @@
...
 Define your APIs using [Smithy](https://smithy.io/2.0/) or [OpenAPI v3](https://swagger.io/specification/), and leverage the power of generated client and server types, infrastructure, documentation, and automatic input validation!
 
-This package vends a projen project type which allows you to define an API using either [Smithy](https://smithy.io/2.0/) or [OpenAPI v3](https://swagger.io/specification/), and a construct which manages deploying this API in API Gateway, given an integration (eg a lambda) for every operation.
+The _type-safe-api_ package provides a projen project type which allows you to define an API using either [Smithy](https://smithy.io/2.0/) or [OpenAPI v3](https://swagger.io/specification/), and a construct which manages deploying this API in an API Gateway, provided an integration (eg a lambda) exists for every operation.
```
cogwirrel: Perhaps we get rid of the ", provided an integration (eg a lambda) exists for every operation" here? Integrations are explained later so it might just add confusion :)

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
```diff
@@ -1,14 +1,19 @@
...
 
-The project will generate "runtime" projects from your API definition in your desired languages, which can be utilised both client side for interacting with your API, or server side for implementing your API. The project also generates a type-safe CDK construct which ensures an integration is provided for every API operation.
+## How does it work?
 
-Code is generated at build time, so when you change your API model, just rebuild and you'll see your changes reflected in the generated code.
+The project will generate runtime projects from your API definition in your desired languages, and you can use both client side for interacting with your API, or server side for implementing your API. The project also generates a type-safe CDK construct which ensures an integration is provided for every API operation.
```
cogwirrel: "...in your desired languages, _which_ you can use..."?

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
```diff
@@ -33,31 +38,47 @@ The `TypeSafeApiProject` projen project will create the following directory stru
...
 ```
 
-Next, you'll need to add `@aws-prototyping-sdk/type-safe-api` to your `NxMonorepoProject`'s `devDeps` and re-synthesize to install the dependency (eg `yarn projen`).
+2. Add `@aws-prototyping-sdk/type-safe-api` to your `NxMonorepoProject`'s `devDeps` and re-synthesize to install the dependency.
+
+```
+yarn projen
+```
+
+The following code shows an example `.projenrc` using `TypeSafeApiProject`.
```
cogwirrel: Might make sense to make this an explicit numbered step, eg "Edit your `.projenrc` and configure TypeSafeApi"?

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
```diff
@@ -358,15 +379,20 @@ See below for an example `.projenrc` making use of `TypeSafeApiProject`. Each ta
     monorepo.synth();
     ```
 
-After defining your `.projenrc`, you'll need to run `projen` and `build` using the appropriate command for your package manager (eg. `yarn projen && yarn build`).
+3. After you define your `.projenrc`, run `projen` and `build` (using the appropriate commands) for your package manager. For example, if you are using `yarn`, use these commands.
```
cogwirrel: I think "for your package manager" should also be in the brackets, since the commands are different for different package managers :)

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
```diff
@@ -499,15 +525,15 @@ In your CDK application, consume the `Api` construct, which is vended from the g
...
-## Implement a Lambda Handler
+## Implement a Lambda handler
 
-The generated runtime projects include lambda handler wrappers which provide type-safety for implementing your API operations. You can implement your lambda handlers in any of the supported languages, and even mix and match languages for different operations if you like.
+The generated runtime projects include lambda handler wrappers which provide type-safety for implementing your API operations. You can implement your lambda handlers in any of the supported languages, or mix and match languages for different operations if you prefer.
 
 === "TS"
 
-    In the above TypeScript CDK application, we used `NodejsFunction` with the entry point as `say-hello.ts`, so we can define the lambda function in the same `infra` project.
+    In the TypeScript CDK application in the CDK construct, we used `NodejsFunction` with the entry point as `say-hello.ts`, so define the lambda function in the same `infra` project.
```
cogwirrel: I think "in the CDK construct" is a bit confusing here :) Also the intention of the "we can" is more "it's possible to" since we're using a NodejsFunction, rather than creating a separate package for the lambda function code :). How about:

"In the TypeScript CDK application, we used `NodejsFunction` which allows you to write your lambda handler code in the same `infra` project. Define `say-hello.ts` and make use of the generated lambda handler wrapper."

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
```diff
@@ -653,24 +679,29 @@ Add the new operation in the `model` project, for example:
...
+
+```ts
 TSError: ⨯ Unable to compile TypeScript:
-src/main.ts(16,7): error TS2741: Property 'sayGoodbye' is missing in type '{ sayHello: { integration: Integration; }; }' but required in type 'OperationConfig<TypeSafeApiIntegration>'.
+src/main.ts(16,7): error TS2741: Property 'sayGoodbye' is missing in type '{ sayHello: { integration: Integration; }; }' but required in type 'Operation
+Config<TypeSafeApiIntegration>'.
 ```
 
-This is because we need to define an integration for every operation in our API. Let's do this next!
+To troubleshoot this, we need to define an integration for every operation in our API.
```
cogwirrel: I'm not sure "to troubleshoot this" makes sense here - the error is expected because the CDK construct requires an integration for every operation, and we just added a new operation which we haven't added an integration for yet :)

Maybe it makes more sense to say:

> Since an integration must be defined for every operation, you may see a build error in your CDK application.

{{error}}

> This is expected, and can be resolved by continuing with the steps below."

cogwirrel (commented):
Thanks for going through this Swapnil! Added a few comments :)

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
swap-aws: @cogwirrel - no, this was me trying to use conventional commits inside markdown. 😬

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
swap-aws: > "...in your desired languages, _which_ you can use..."?

updated

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
swap-aws: > Perhaps we get rid of the ", provided an integration (eg a lambda) exists for every operation" here? Integrations are explained later so it might just add confusion :)

removed

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
swap-aws: updated

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
swap-aws: updated

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
swap-aws: updated

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
swap-aws: updated

swap-aws (commented):
Updated from feedback

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
cogwirrel: ooh we probably still want to delete this before we merge :) otherwise looks great! 😄

cogwirrel (approved):

packages/type-safe-api/docs/developer_guides/type-safe-api/index.md
```diff
@@ -1,14 +1,18 @@
...
+You can define your APIs using [Smithy](https://smithy.io/2.0/) or [OpenAPI v3](https://swagger.io/specification/), and leverage the power of generated client and server types, infrastructure, documentation, and automatic input validation.
 
-The project will generate "runtime" projects from your API definition in your desired languages, which can be utilised both client side for interacting with your API, or server side for implementing your API. The project also generates a type-safe CDK construct which ensures an integration is provided for every API operation.
+## How does it work?
 
-Code is generated at build time, so when you change your API model, just rebuild and you'll see your changes reflected in the generated code.
+The project will generate runtime projects from your API definition in your desired languages, which you can use both on the client side for interacting with your API, or server side for implementing your API. The project also generates a type-safe CDK construct which ensures an integration is provided for every API operation.
 
-The `TypeSafeApiProject` projen project will create the following directory structure within its `outdir`:
+!!! note
```
cogwirrel, line 11: Ooh do you know if this (and the other `!!!` directives) will render differently in the UI? Might be worth building locally and taking a look at the generated html documentation if you haven't already :)

cogwirrel:
Looks great! I think the only thing would be to check that the `!!!note` etc syntax works ok for the docs website, and if not perhaps consider a different way to represent them? Otherwise happy to ship! :)

swap-aws:
@cogwirrel, @agdimech tried it out and it works fine :)

cogwirrel:
> @cogwirrel, @agdimech tried it out and it works fine :)

Fantastic! 😄

cogwirrel (approved):

swap-aws:
@cogwirrel I don't think I have merge rights, so feel free to merge if needed.

PR #477, opened 2023-06-21
chore(type-safe-api): more verbose scripts by default

cogwirrel:
The `TYPE_SAFE_API_DEBUG=1` flag is now just used for the extra logs. Failures to run openapi-generator or other scripts are no longer suppressed by default, so that users can get more actionable error messages than "script exited with code 1".

A recent example was a customer found the `generate` command was failing due to not having Java installed, but could only discover that this was the root cause by adding the `TYPE_SAFE_API_DEBUG=1` flag.

### DIFF
--- a/packages/type-safe-api/scripts/common/common.sh
+++ b/packages/type-safe-api/scripts/common/common.sh
@@ -3,12 +3,9 @@
 set -e
 
 ##
-# log debug messages if TYPE_SAFE_API_DEBUG env var set to 1
-# this is a helper function for debugging our scripts
+# log messages
 log() {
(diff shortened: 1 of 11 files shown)


agdimech (approved):
There are a few other places that are using this flag. Can you clean these up also:

https://github.com/search?q=repo%3Aaws%2Faws-prototyping-sdk%20TYPE_SAFE_API_DEBUG&type=code

agdimech (dismissed):
Please remove references here: https://github.com/search?q=repo%3Aaws%2Faws-prototyping-sdk%20TYPE_SAFE_API_DEBUG&type=code

agdimech (approved):

PR #489, opened 2023-07-03
feat(static-website): add bucketdeployment options

### LINKED ISSUE #480: [FEATURE] Add BucketDeploymentProps in StaticWebsite
### Describe the feature

StaticWebsite is almost **perfect** and allow for high customization but one thing is missing: `BucketDeploymentProps`

### Use Case

I would like to add customization for `BucketDeployment` for example cache control and log retention

### Proposed Solution

We could easily add customization by adding `bucketDeploymentProps` like the `distributionProps` in `distribution`

```ts
    // Deploy Website
    this.bucketDeployment = new BucketDeployment(this, "WebsiteDeployment", {
      // Add following line
      ...bucketDeploymentProps,

      // Existing code
      sources: [
        Source.asset(props.websiteContentPath),
        ...(props.runtimeOptions
          ? [
              Source.jsonData(
                props.runtimeOptions?.jsonFileName ||
                  DEFAULT_RUNTIME_CONFIG_FILENAME,
                props.runtimeOptions?.jsonPayload
              ),
            ]
          : []),
      ],
      destinationBucket: this.websiteBucket,
      // Files in the distribution's edge caches will be invalidated after files are uploaded to the destination bucket.
      distribution: this.cloudFrontDistribution,
    });
```

And of course we need to add the following in StaticWebsiteProps

```ts
  readonly bucketDeploymentProps?: BucketDeploymentProps;
```

### Other Information

_No response_

### Acknowledgements

- [x] I may be able to implement this feature request

### PDK version used

0.19.27

Typescript

macOS Apple M1 Pro, Ventura 13.4.1

agdimech:
fix #480

### DIFF
--- a/packages/static-website/src/static-website.ts
+++ b/packages/static-website/src/static-website.ts
@@ -11,9 +11,13 @@ import {
   OriginBindOptions,
   ViewerProtocolPolicy,
 } from "aws-cdk-lib/aws-cloudfront";
+import * as cloudfront from "aws-cdk-lib/aws-cloudfront";
 import { S3Origin } from "aws-cdk-lib/aws-cloudfront-origins";
+import * as ec2 from "aws-cdk-lib/aws-ec2";
 import { PolicyStatement } from "aws-cdk-lib/aws-iam";
+import * as iam from "aws-cdk-lib/aws-iam";
 import { Key } from "aws-cdk-lib/aws-kms";
+import * as logs from "aws-cdk-lib/aws-logs";
 import {
   BlockPublicAccess,
   Bucket,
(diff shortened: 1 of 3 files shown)


packages/static-website/src/static-website.ts
```diff
@@ -102,6 +106,20 @@ export interface StaticWebsiteProps {
...
   readonly distributionProps?: DistributionProps;
 
+  /**
+   * Custom bucket deployment properties.
+   *
+   * NOTE: sources and destinationBucket are required, however are overriden by construct specific instances. Please pass in the following as a minimum:
+   * ```ts
+   * {
+   *   sources: [],
+   *   // @ts-ignore
```
mteichtahl: nitt - not sure this is a good example. perhaps update with a more "real-life" example - ie: there should be no need for the // @ts-ignore in the example

mteichtahl (changes requested):
just one little nitt

packages/static-website/src/static-website.ts
agdimech: This is actually the way i'd want people to use it as otherwise they would need to create a bucket which doesn't get used. This is the problem with JSII atm as there is no way to do Partial without having to re-implement the props ourselves.

agdimech: On second thought, I have just forked `BucketDeploymentProps` and removed the required fields to make this simpler to use. Downside is more boilerplate code :(

mteichtahl (approved):

PR #502, opened 2023-07-12
docs: reviewed Integrations topic

swap-aws:
Fixes #


packages/type-safe-api/docs/developer_guides/type-safe-api/integrations.md
```diff
@@ -45,20 +43,26 @@ For example:
     )
     ```
 
-## Lambda Integration
+## Lambda integration
+
+For integrating an API operation with a lambda, use the `Integrations.lambda(yourLambdaFunction)`.
```
cogwirrel: Nit: let's get rid of `the` :)

packages/type-safe-api/docs/developer_guides/type-safe-api/integrations.md
```diff
@@ -45,20 +43,26 @@ For example:
...
+
+## Mock integration
+
+To mock an API operation, use the `MockIntegrations` class, which is included in your generated infrastructure package. This contains an integration for every response that can be returned by your operations.
 
-For integrating an API operation with a lambda, you can use the `Integrations.lambda(yourLambdaFunction)`.
+!!! info
 
-## Mock Integration
+    If you are using `TypeSafeApi` directly or need to mock a response not defined in your model, use `Integrations.mock` for a less type-safe alternative.
```
cogwirrel: Maybe let's just say "If you need to mock a response not defined in your model, use `Integrations.mock` for a less type-safe alternative."

Using the `TypeSafeRestApi` construct directly would be a pretty rare case and I don't think we need to call it out in the docs :)

cogwirrel (commented):
Nice work! Sorry I think both my comments are on things I wrote - this is a good chance for me to review the mistakes I made on the first draft I guess! 🤣

cogwirrel (approved):

PR #504, opened 2023-07-12
fix(type-safe-api): copy openapi spec into infra package so that it gets published

zsstiers:
Moving the generated openapi spec allows it to show up in `npm pack` and by extension get published as part of the package. This also includes adding minimal .npmignore files because the default behavior of `npm pack` leads to inheriting potentially problematic ignores from higher up the package hierarchy in monorepos. To avoid potential conflicts between projen and openapi generator the default .npmignore created by projen is not used and is instead replaced with a tiny safer ignore file.


cogwirrel (approved):
Awesome, thank you! 🚀🚀

PR #513, opened 2023-07-25
fix: dynamically generate website content based on presence of API

agdimech:
Add support for passing in a `TypeSafeApiProject` into the `CloudscapeReactTsWebsiteProject`. If this is passed in, the website will generate sample code assuming API integration is required, and will configure dependencies, hooks and sample page (Home) to call the API.

Note: This assumes the API contains a `sayHello` API which would be the case if it was newly created. If the API no longer    has this operation, the website will fail to build.


packages/cloudscape-react-ts-website/templates/src/components/DefaultApi/index.tsx.mustache
```diff
@@ -0,0 +1,25 @@
+/*! Copyright [Amazon.com](http://amazon.com/), Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0 */
+import { DefaultApiClientProvider as DefaultApiClientProvider_ } from "myapi-typescript-react-query-hooks";
```
cogwirrel: Missing the package placeholder here :)

cogwirrel (approved):

packages/cloudscape-react-ts-website/templates/src/components/DefaultApi/index.tsx.mustache
agdimech: Great find - fixed!

PR #515, opened 2023-07-26
fix(type-safe-api): reduce strictness of ts nx monorepo parent check

cogwirrel:
instanceof will return false if the version of nx-monorepo the user has installed does not exactly match the one type-safe-api depends on, so we fall back to check that the parent looks like the typescript nx-monorepo.

### DIFF
--- a/packages/type-safe-api/src/project/type-safe-api-project.ts
+++ b/packages/type-safe-api/src/project/type-safe-api-project.ts
@@ -166,7 +166,9 @@ export class TypeSafeApiProject extends Project {
     super(options);
 
     const nxWorkspace = this.getNxWorkspace(options);
-    const isNxTsWorkspace = this.parent instanceof NxMonorepoProject;
+
+    const isNxTsWorkspace =
+      this.parent && this.parent.constructor.name === NxMonorepoProject.name;
 
     // API Definition project containing the model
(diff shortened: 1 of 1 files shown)


packages/type-safe-api/src/project/type-safe-api-project.ts
```diff
@@ -166,7 +166,15 @@ export class TypeSafeApiProject extends Project {
     super(options);
 
     const nxWorkspace = this.getNxWorkspace(options);
-    const isNxTsWorkspace = this.parent instanceof NxMonorepoProject;
+
+    // instanceof will return false if the version of nx-monorepo the user has installed does not exactly match the one type-safe-api depends on,
+    // so we fall back to check that the parent looks like the typescript nx-monorepo.
+    const isNxTsWorkspace =
```
agdimech, line 170: Just wandering if we could do this instead?

`const isNxTsWorkspace = (<any>this.parent).constructor.name === NxMonorepoProject.name`

cogwirrel: Yeah I think that's cleaner and should work! Will update :)

cogwirrel:
Closing in favour of https://github.com/aws/aws-prototyping-sdk/pull/517

PR #516, opened 2023-07-26
fix: resolve issues with typesafe dep from cloudscape

agdimech:
- Use `file:` dep when linking to type-safe api to ensure this works in non-typescript monorepos (workspaces won't be enabled).
- Ensure root level tasks invoke `nx run-many` for all languages


JoshuaToth (approved):
lgtm

packages/cloudscape-react-ts-website/src/cloudscape-react-ts-website-project.ts
```diff
@@ -74,15 +74,19 @@ export class CloudscapeReactTsWebsiteProject extends ReactTypeScriptProject {
...
           "Cannot pass in a Type Safe Api without React Hooks Library configured!"
         );
       }
-      this.addDeps(libraryHooksPackage);
+      this.addDeps(
+        `${libraryHooksPackage}@file:${path.relative(
+          this.outdir,
+          hooks.outdir
+        )}`
+      );
```
cogwirrel, line 89: Maybe we should only do this when the parent isn't a ts monorepo? I know file:// will always work but it feels cleaner not to force the file dep in the ts monorepo case :)

cogwirrel (approved):

packages/cloudscape-react-ts-website/src/cloudscape-react-ts-website-project.ts
agdimech: will keep as is for now. We can potentially add it as part of your ongoing PR :)

PR #528, opened 2023-08-09
fix(static-website): allow for multiple cloudfront web acls in a single stack

cogwirrel:
The custom resource lambda name is unique per stack, but created for every instance of `CloudfrontWebAcl`, meaning that multiple instances of `CloudfrontWebAcl` will conflict with each other if used within the same stack. This change adds part of the unique construct address hash to the name to ensure it is unique enough to use multiple instances in the same stack.

### DIFF
--- a/packages/static-website/src/cloudfront-web-acl.ts
+++ b/packages/static-website/src/cloudfront-web-acl.ts
@@ -116,9 +116,6 @@ export class CloudfrontWebAcl extends Construct {
    * @private
    */
   private createOnEventHandler(stack: Stack, aclName: string): Function {
-    const onEventHandlerName = `${PDKNag.getStackPrefix(stack)
-      .split("/")
-      .join("-")}OnEventHandler`;
     const onEventHandlerRole = new Role(this, "OnEventHandlerRole", {
       assumedBy: new ServicePrincipal("lambda.amazonaws.com"),
       inlinePolicies: {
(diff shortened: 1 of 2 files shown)


packages/static-website/src/cloudfront-web-acl.ts
```diff
@@ -118,7 +118,7 @@ export class CloudfrontWebAcl extends Construct {
   private createOnEventHandler(stack: Stack, aclName: string): Function {
     const onEventHandlerName = `${PDKNag.getStackPrefix(stack)
       .split("/")
-      .join("-")}OnEventHandler`;
+      .join("-")}OnEventHandler-${this.node.addr.slice(0, 4)}`;
```
JeremyJonas: Why can't we just use dynamic name for this? What is reason to use explicit naming?

JeremyJonas (approved):
Approving conditionally - if we absolutely need explicit name here, then ok - but would prefer to use dynamic name when/where ever possible. This code looks like can be dynamic.

packages/static-website/src/cloudfront-web-acl.ts
cogwirrel: Looks like it was explicit so that it could be used for a more restrictive policy for writing to cloudwatch logs. I don't think there's any harm in moving the wildcard to the `/aws/lambda/*` log prefix though, so will update to a dynamic name :)

PR #538, opened 2023-08-18
feat(type-safe-api): add compilation for ESM modules

zetashift:
That commitizen workflow is new! :P


packages/type-safe-api/src/project/codegen/runtime/generated-typescript-runtime-project.ts
```diff
@@ -123,6 +123,20 @@ export class GeneratedTypescriptRuntimeProject extends TypeScriptProject {
...
+    new JsonFile(this, "tsconfig.esm.json", {
+      obj: {
+        extends: "./tsconfig.json",
+        compilerOptions: {
+          module: "esnext",
+          moduleResolution: "nodenext",
+          outDir: "lib/esm",
+        },
+      },
+    });
```
zetashift: how do I conditionally add this generation to the task? This file and the fields in `package.json` need to only be added when this task is run, right?

.husky/pre-commit
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -eo pipefail
+# set -eo pipefail
```
zetashift: I'm using Ubuntu 22.10, I have no idea why this makes `pnpm cz` fail:
```
? Does this change affect any open issues? No
.husky/pre-commit: 2: set: Illegal option -o pipefail
git exited with error code 1
```

AFAIK `-o pipefail` is not an illegal option?

packages/type-safe-api/src/project/codegen/runtime/generated-typescript-runtime-project.ts
agdimech: We probably should be extending `tsconfig.dev.json `as this is what we actually use to compile.

There is no harm in having this file generated and the task instrumented all the time.

also, you probably want to execute this during compile, to do that you can add this after your task definition:

`this.compileTask.spawn(compileESMTask);`

.husky/pre-commit
agdimech: It’s because in ubtunu `sh` is symlinked to `dash` which doesn’t support this option. Fix should be to replace the shebang on the first line with `#!/bin/bash`

agdimech (commented):
Great start - thanks for contributing.

.husky/pre-commit
zetashift: This didn't work for me, I got the  same error. I also tried `#!/usr/bin/bash` and other variants.

This works:
```
#!/bin/bash
set -e pipefail
```
According to a quick google this still fails early in pipelines, but I'm not sure. I can try and open up a new PR instead on my Mac, where this commit should just work (tm).

packages/type-safe-api/src/project/codegen/runtime/generated-typescript-runtime-project.ts
cogwirrel: I think we want to keep a `tsconfig.esm.json` that's separate so we can build both cjs and esm :)

cogwirrel:
Looks great! Thanks for this! :D

zetashift:
Since all the husky changes and conflicts I'll just open a new PR!

Sorry for the poor Git-fu! :P

PR #542, opened 2023-08-21
feat(type-safe-api): add default interceptors for typescript java and python handlers

cogwirrel:
Add 5 default interceptors - TryCatchInterceptor for error handling, CorsInterceptor for adding cors headers, and Metrics, Tracing and Logging using AWS Powertools for Lambda. Use these interceptors by default in the generated handler stubs. Additionally adds a "Response" class for TypeScript and Python for simplifying construction of the response objects. Note that TryCatchInterceptor will catch and return any api responses that are thrown as exceptions.


agdimech (approved):
NOICE

PR #544, opened 2023-08-23
feat(pipeline): enable setting CodeBuild properties on FeatureBranch projects

jstrunk:
Apply the properties from the PDKPipeline `codeBuildDefaults` property to the CodeBuild project that creates feature branch pipelines. This enables  configuration such as role policies, filesystems, and networking. The role policy is useful to enable access to private repos in CodeArtifact.

The feature branch CodeBuild project now defaults to the `STANDARD_7_0` build image.

References #350


agdimech:
HI Jeff - this looks good although we should not mark this commit as a BREAKING CHANGE as we should only reserve this to when a backward-incompatible API change occurs. Please update your commit accordingly then happy to merge.

As always, thanks for your contribution!

PR #557, opened 2023-09-03
docs(type-safe-api): update docs and migration guide

cogwirrel:
Also enable push to github pages, and add missing webAclOptions property to generated java cdk construct.

Fixes #


agdimech (approved):
Nice

PR #559, opened 2023-09-03
docs: update migration guide

agdimech:
Fixes #


docs/content/getting_started/migration_guide.md
```diff
@@ -630,3 +669,26 @@ In AWS PDK, these classes have been moved out from `Handlers` as their own stand
         }
     }
     ```
+
+## Troubleshooting
+### Missing Polyfill
+
+If you see this error:
+
+```
```
cogwirrel, line 398: Probably best not to make Greg famous 😉

cogwirrel: Also, this can be caused by using the ts runtime in a website- might be worth mentioning the hooks lib should be used in a website even if you just use the generated client and not the hooks, since the runtime has all the server side handlers so is less suitable for web.

agdimech: Good catch - resolved and added section regarding website -> runtime.

cogwirrel (approved):

PR #563, opened 2023-09-05
fix(type-safe-api): ensure lambda permission policy size does not exceed limit for router

### LINKED ISSUE #549: [BUG] TypeSafeApi - resource-based policies exceed hard limit of 20K
### Describe the bug

For an API with a large surface area (>50 operations), TypeSafeApi generates a lambda resource policy that exceeds the AWS hard limit (according to ```https://docs.aws.amazon.com/lambda/latest/dg/access-control-resource-based.html```) which causes deployment failures.

### Expected Behavior

The generated resource policy should not exceed 20Kb and allow effectively unlimited operations/handlers

### Current Behavior

The implementation begin used is based on ```handlerRouter```

```
export const handler = handlerRouter({
  interceptors: [corsInterceptor, tryCatchInterceptor, identityInterceptor],
  handlers: { ... } // 51 handlers
});
```

on deployment, cloudFormation throws
```
Resource handler returned message: "The final policy size (20778) is bigger than the limit (20480). (Service: Lambda, Status Code: 400, Request ID: 3fbf0839-1916-4983-90c0-b53a7ddd2ce0)" (RequestToken: 44ce75cc-79b2-9de4-5d28-0ff14066ab80, HandlerErrorCode: InvalidRequest)
```

### Reproduction Steps

```
export const handler = handlerRouter({
  interceptors: [corsInterceptor, tryCatchInterceptor, identityInterceptor],
  handlers: { ... } // 51 handlers
});
```

### Possible Solution

Optimisation of ```https://github.com/aws/aws-prototyping-sdk/blob/1fcf948a2f0f87ee28f69e231f79a3e8ce6b6ea6/packages/type-safe-api/src/construct/integrations/lambda.ts#L46C4-L57C14``` when ```handlerRouter``` is being used

_No response_

### PDK version used

0.19.47

Typescript

Darwin 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000 arm64

cogwirrel:
When using the handler router (a single lambda function for all operations), type safe api would still create a new permission to grant api gateway permissions to invoke the lambda for each individual operation.

With this change, we retain the granular permissions for individual lambdas, but as soon as a lambda function is used for multiple operations, we replace the granular permission with a broader permission for api gateway to invoke it for any operation, so as not to exceed the maximum policy size.

Fixes #549

### DIFF
--- a/packages/type-safe-api/src/construct/integrations/integration.ts
+++ b/packages/type-safe-api/src/construct/integrations/integration.ts
@@ -2,7 +2,11 @@
 SPDX-License-Identifier: Apache-2.0 */
 import { SpecRestApi } from "aws-cdk-lib/aws-apigateway";
 import { Construct } from "constructs";
-import { OperationDetails, SerializedCorsOptions } from "../spec";
+import {
+  OperationDetails,
+  OperationLookup,
+  SerializedCorsOptions,
+} from "../spec";
 
 /**
  * Specifies an API method integration type
@@ -183,6 +187,10 @@ export interface IntegrationRenderProps extends OperationDetails {
    * Cross Origin Resource Sharing options for the API
    */
   readonly corsOptions?: SerializedCorsOptions;
+  /**
+   * Details about all operations in the API
+   */
+  readonly operationLookup: OperationLookup;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


agdimech (approved):
Approved, assuming tests will be fixed :)

PR #575, opened 2023-09-13
fix(type-safe-api): revert add compilation to ESM for TypeScript

cogwirrel:
Reverts aws/aws-pdk#561

Unfortunately, it seems as though this combination of `module` and `moduleResolution` no longer works as of TypeScript 5.2.

With TS 5.2, we get the error:

```
tsconfig.esm.json(4,15): error TS5110: Option 'module' must be set to 'NodeNext' when option 'moduleResolution' is set to 'NodeNext'.
```
See: https://www.totaltypescript.com/concepts/option-module-must-be-set-to-nodenext-when-option-moduleresolution-is-set-to-nodenext#solution-2

It seems as though with TS 5.2 it's tricky to build hybrid esm/cjs packages: https://github.com/microsoft/TypeScript/issues/55705

So unfortunately reverting this as the path of least resistance for now. Sorry @zetashift ! :(


agdimech (approved):

zetashift:
Ha, classic CJS vs ESM! Well I do love the path of least resistance :P, so all good.

PR #582, opened 2023-09-22
docs: use northstar infinite query table in shopping list guide

cogwirrel:
Adds table pagination to the guide without any extra code


agdimech (approved):
Nice - this is awesome

PR #585, opened 2023-09-27
fix(type-safe-api): omit pyproject toml from python runtime project code generation

cogwirrel:
`pyproject.toml` is managed by projen, and openapi generator should not overwrite this


JeremyJonas (approved):
Thank you!

PR #587, opened 2023-09-29
fix(monorepo): ensure local venv is used for all subproject tasks

cogwirrel:
The poetry env info command will print the parent environment if the `VIRTUAL_ENV` environment variable is already set to the parent environment. Ensure this gets reset for all subproject tasks to address issues where dependencies would appear to be missing during build.


agdimech (approved):
This looks good :)

PR #589, opened 2023-10-01
fix(docs): fix migration dependency for @aws/pdk/monorepo (not nx-monorepo)

### LINKED ISSUE #588: [DOCS] (Migration Step 2 references @aws/pdk/nx-monorepo incorrectly)
### Describe the issue

Step 2 "diff" lists `+import { MonorepoTsProject } from "@aws/pdk/nx-monorepo";` as change which is not correct.  needs to be `+import { MonorepoTsProject } from "@aws/pdk/monorepo";`.

### Links

https://aws.github.io/aws-pdk/getting_started/migration_guide.html

walmsles:
Fixes #588

Update migration `diff` in step 2 of the migration to `@aws/pdk/monorepo`.  The existing diff still has the change as `@aws/pdk/nx-monorepo` which is incorrect.


agdimech (approved):
Great catch - thank you so much!

PR #592, opened 2023-10-05
feat(type-safe-api): add s3 integration

v41e:
As discussed here https://github.com/aws/aws-pdk/discussions/539, this is a proposition to add an s3 integration to type-safe-api


packages/type-safe-api/src/construct/integrations/integrations.ts
```diff
@@ -23,4 +24,12 @@ export class Integrations {
   public static mock(response: MockIntegrationResponse): MockIntegration {
     return new MockIntegration(response);
   }
+
+  /**
+   * An integration that can read/write to an S3 bucket
```
cogwirrel, line 29: Would be great to link these docs here too :) https://docs.aws.amazon.com/apigateway/latest/developerguide/integrating-api-with-aws-services-s3.html

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.pathOverride = props.pathOverride;
+  }
+
+  /**
+   * Render the lambda integration as a snippet of OpenAPI
+   */
+  public render(props: IntegrationRenderProps): ApiGatewayIntegration {
+    return {
+      type: "AWS",
+      httpMethod: props.method.toUpperCase(),
```
cogwirrel: It might be useful to expose a `method` option to override the method? (ideally nicely typed so users know they need to specify it in upper case)

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,89 @@
...
+   * @default - a new role will be created
+   */
+  readonly role?: IRole;
+
+  /**
+   * The path override for the integration
+   * @default - integration path is used
+   * @see https://docs.aws.amazon.com/apigateway/latest/developerguide/integrating-api-with-aws-services-s3.html
+   */
+  readonly pathOverride?: string;
```
cogwirrel: Perhaps we just name this `path` and given it's optional it still makes sense as an override :)

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    if (props.role) {
+      this.role = props.role;
+    } else {
+      this.role = new Role(this.bucket, "Role", {
+        assumedBy: new ServicePrincipal("apigateway.amazonaws.com"),
+      });
+    }
+
+    this.bucket.grantReadWrite(this.role, this.pathOverride ?? "*");
```
cogwirrel: It would be best to scope this down if possible :)

Can we move the grant into an override of the `grant` method, like this:

https://github.com/aws/aws-pdk/blob/917a5db21d96d620ade7f306f9fedbca16ad1648/packages/type-safe-api/src/construct/integrations/lambda.ts#L43

This way we can access the `method` to determine whether to grant read or write permissions, and the `path` to determine the key pattern :)

Note we need to be careful with path parameters (eg `/foo/{myParam}`) - notice how we replace those with `*` in the lambda grants:

https://github.com/aws/aws-pdk/blob/917a5db21d96d620ade7f306f9fedbca16ad1648/packages/type-safe-api/src/construct/integrations/lambda.ts#L100-L105

packages/type-safe-api/src/construct/spec/utils.ts
```diff
@@ -16,3 +17,19 @@ export const functionInvocationUri = (
...
+ * @param scope scope in which the s3 is deployed
+ * @param bucket the s3 bucket to be invoked
+ */
+export const bucketInvocationUri = (
+  scope: Construct,
+  bucket: IBucket,
+  pathOverride?: string
+): string => {
+  const stack = Stack.of(scope);
+  return `arn:${stack.partition}:apigateway:${stack.region}:s3:path/${
```
cogwirrel: I can't find the relevant documentation re cross-region s3 integrations - is it definitely the case that we want to use the API's region in this uri (from the `scope` variable) rather than the region for the bucket?

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,89 @@
...
+      httpMethod: props.method.toUpperCase(),
+      uri: bucketInvocationUri(
+        props.scope,
+        this.bucket,
+        this.pathOverride ?? props.path
+      ),
+      credentials: this.role.roleArn,
+      responses: {
+        default: {
+          statusCode: "200",
```
cogwirrel: I had a quick play around with this integration and noticed if an object is missing in the bucket I get back some xml saying it's not found but I get a `200` response - is that just the way it is for S3 integrations (I haven't used them before!), or is there something we can do to make it respond with a `404` in this case? :)

The ideal behaviour I think would be to map the non 200 responses to the shape of the error responses defined in the model - eg in the model if there's a 404 response which is json with a "message" property, ideally the message back from s3 should be injected and a json response returned... I can see there being a lot of cases to manage though so probably out of scope for now - but if it's possible to return json for errors like 404s instead of xml it would at least play a bit nicer with the generated clients when invoking these apis.

cogwirrel (commented):
Thanks so much for this!

I've added a few comments below :) Also please could you add some documentation [here](https://github.com/aws/aws-pdk/blob/917a5db21d96d620ade7f306f9fedbca16ad1648/packages/type-safe-api/docs/developer_guides/type-safe-api/integrations.md?plain=1#L49)? Would be great to include some examples in the docs of how to read/write some simple json objects :)

This is a really great start! I can see this being a really useful addition to type-safe-api - thanks a bunch! 😄

v41e:
Sure, I totally forgot the documentation, I'll update it too

packages/type-safe-api/src/construct/spec/utils.ts
v41e: My bad you're right, there is not really a documentation about it, but if you try to configure an API like so in the console and export the API with API gateway extension you'll see that's the bucket region and so i supposed the bucket partition too.

You should probably update the `functionInvocationUri` too because it's the lambda region and partition too. I know you generate handlers in same region but in case a user might want to use a custom handler

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: That's the tricky part because S3 can send a lot of errors.

Maybe a way to handle it is to provide different implementations:
- one that do nothing
- one that simply aggregate all 4XX and all 5XX to a simple error
- one that totally customizable by the user

packages/type-safe-api/src/construct/spec/utils.ts
cogwirrel: Ah nice one! Thanks for looking into this - I'll fix the lambdas! 😄

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: Nice idea! Let's go with that approach if it's not too hard to build :) If the totally customizable one is too difficult to implement we can skip it, as users can always define a custom integration themselves :)

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,89 @@
...
+          responseTemplates: {},
+        },
+        "4|5\\d{2}": {
+          statusCode: "500", // force 500 because 4XX errors is not a real client error, it's a misconfiguration on the server side
+          responseParameters: props.corsOptions
+            ? generateCorsResponseParameters(props.corsOptions)
+            : {},
+          responseTemplates: {},
+        },
+      },
```
cogwirrel, line 114: Ooh sorry I think one more thing we should probably support is path parameters, like in that example where the path parameters can be used in the s3 key from the docs: https://docs.aws.amazon.com/apigateway/latest/developerguide/integrating-api-with-aws-services-s3.html#api-items-in-folder-as-s3-objects-in-bucket

I think customising the bucket is definitely out of scope, but perhaps we just support a simple mapping where the path parameters are all included in the integration path parameters and so used for variable keys 😄 

I think to achieve this all you'd need to do is add the same thing as I did for the custom nlb integration to `requestParameters`:

https://github.com/aws/aws-pdk/blob/917a5db21d96d620ade7f306f9fedbca16ad1648/packages/type-safe-api/docs/developer_guides/type-safe-api/custom_integration_ecs.md?plain=1#L242-L249

I think we'd also get the ability to use them differently in the overridden path for free which is also a pretty cool feature! :) eg the api might be `/pets/{petId}` which might map to `<bucket>/pet-data/{petId}/details.json` or something :)

cogwirrel:
@valebedu just wondered if you're still working on this? Would be an awesome addition but no worries if you don't have the time! :)

v41e:
Hi @cogwirrel, yes it's just a bit difficult to find time 😅 I'll try to work on it this weekend

cogwirrel:
> Hi @cogwirrel, yes it's just a bit difficult to find time 😅 I'll try to work on it this weekend

All good, no rush! :) Will keep this PR alive 😄

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: @cogwirrel what do you think of the proposed `ErrorIntegrationResponse`?

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    if (props.role) {
+      this.role = props.role;
+    } else {
+      this.role = new Role(this.bucket, "Role", {
+        assumedBy: new ServicePrincipal("apigateway.amazonaws.com"),
+      });
+    }
+
+    this.bucket.grantReadWrite(this.role, this.path ?? "*");
```
cogwirrel: Looks like `this.path` is referenced before it's set :)

Also if the path has parameters in it we'll need to replace them with wildcards here, eg:

`pets/{petId}/details.json` becomes `pets/*/details.json`

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: I think it's a really neat way to do it - thanks so much! :)

packages/type-safe-api/test/construct/type-safe-rest-api.test.ts
```diff
@@ -299,6 +300,48 @@ describe("Type Safe Rest Api Construct Unit Tests", () => {
...
+              bucket: new Bucket(stack, "Bucket", {}),
+            }),
+          },
+        },
+      });
+      expect(Template.fromStack(stack).toJSON()).toMatchSnapshot();
+      snapshotExtendedSpec(api);
+    });
+  });
+
```
cogwirrel, line 346: Please could we add a test for each of the error integration response types? :)

packages/type-safe-api/src/construct/integrations/error-integration-response.ts
```diff
@@ -0,0 +1,109 @@
...
+              ...errorResponse.responseParameters,
+              ...generateCorsResponseParameters(props.corsOptions),
+            },
+            contentHandling: errorResponse.contentHandling,
+          };
+        }
+
+        result[responseStatusPattern] = errorResponse;
+      }
+    }
```
cogwirrel: Nit: you can simplify this to something like:

```ts
return Object.fromEntries(Object.entries(this.errorResponses ?? {}).map(([responseStatusPattern, errorResponse]) => [responseStatusPattern, {
  ...errorResponse,
  responseParameters: {
    ...errorResponse.responseParameters,
    ...(props.corsOptions ? generateCorsResponseParameters(props.corsOptions) : {}),
  },
}]));
```

packages/type-safe-api/src/construct/integrations/error-integration-responses.ts
```diff
@@ -0,0 +1,35 @@
...
+   * A custom error integration response where you can specify the error responses as you wish
+   * @param props the integration props
+   * @see https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-integration-settings-integration-response.html
+   */
+  public static custom(
+    props: CustomErrorIntegrationResponseProps
+  ): CustomErrorIntegrationResponse {
+    return new CustomErrorIntegrationResponse(props);
+  }
+}
```
cogwirrel: Nice work!! I really like this approach :)

cogwirrel (commented):
Thanks so much @valebedu! Just a few minor comments :) I think the main thing remaining if you don't mind is to make sure path parameters are mapped appropriately :)

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: Great, and it could be used in other integrations too, but it's totally useless in LambdaIntegration because API Gateway act as a proxy and that's the lambda job to return errors

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: @cogwirrel  I tried something but i'm not really familiar with custom grants and it's a bit more complicated because the render method need the role too because we need to specify the roleArn in credentials. So I'll push something but please note it's a work in progress and I'll need some feedback about it

packages/type-safe-api/test/construct/type-safe-rest-api.test.ts
v41e: Yes that's in my todo 👌

packages/type-safe-api/src/construct/integrations/error-integration-response.ts
v41e: Thanks a lot, this is much cleaner 🙌

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: No worries! :) I think we can have the constructor store/create the role but then grant the permissions in the integration's `grant` method so we have the final `path` and `method` if they're not overridden by the user :)

I was thinking we could do something like this:

Constructor:

```ts
this.role = props.role ?? new Role(this.bucket, "Role", {
  assumedBy: new ServicePrincipal("apigateway.amazonaws.com"),
});
this.method = props.method;
// ...etc
```

Then override the integration `grant` method to do the granting, eg:

```ts
public grant(props: IntegrationGrantProps) {
  const path = this.path ?? props.path;
  const method = this.method ?? props.method;

  // Replace path parameters with * to grant access to arbitrary values for path parameters
  const permissionPath = path?.replace( /{[^\}]*\}/g, "*") ?? "*";

  // Grant read access for GET/HEAD/OPTIONS/TRACE, otherwise read/write
  if (["get", "head", "options", "trace"].includes(method.toLowerCase())) {
    this.bucket.grantRead(this.role, permissionPath);
  } else {
    this.bucket.grantReadWrite(this.role, permissionPath);
  }
}
```

That way we can fall back to using the `path` and `method` defined in the API for the permissions, rather than providing `*` if the user doesn't override the path :)

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: But `this.path` is set in construction from props, what do you mean?

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: Did you see my last commit with my proposal?

v41e: I think I scoped it too much down 😂

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: Ok I see it's outdated

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,179 @@
...
+              // Path parameters (eg {param} are replaced with wildcards)
+              resourceName: `*/${method.toUpperCase()}${path.replace(
+                /{[^\}]*\}/g,
+                "*"
+              )}`,
+            }),
+          },
+        },
+      })
+    );
```
cogwirrel: I'm not sure if this will actually do anything as it's a grant on the role arn with no actions :)

https://github.com/aws/aws-cdk/blob/7c497ee998976b69ad1e2cd3e68f30cb593f8ded/packages/aws-cdk-lib/aws-iam/lib/role.ts#L573-L580

I think you want `grantAssumeRole`? :)

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    const bucketMethod = this.method ?? method;
+
+    var bucketActions: string[] = [];
+    var keyActions: string[] = [];
```
cogwirrel: Nit: Prefer `let` to `var` :)

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,179 @@
...
+        break;
+
+      case "delete":
+        bucketActions = perms.BUCKET_DELETE_ACTIONS;
+        break;
+    }
+
+    Grant.addToPrincipalOrResource({
+      grantee: this.role,
+      actions: bucketActions,
```
cogwirrel: Nice idea having the grant shared but switching on the method for more minimal permissions :)

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,179 @@
...
+      resourceArns: [
+        this.bucket.bucketArn,
+        this.bucket.arnForObjects(this.path ?? path),
```
cogwirrel: We'll still need to swap any path parameters with `*` here, ie:

```
this.bucket.arnForObjects((this.path ?? path ?? "*").replace(/{[^\}]*\}/g, "*"))
```

Which will mean eg `/pets/{petId}/details.json` becomes like `/pets/*/details.json` in the permissions :)

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: I don't understand how it could map something like in the tutorial

```
{bucket}: method.request.path.folder
```

with:
- {bucket}: the path parameter override
- folder: the real path parameter

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,179 @@
...
+    this.bucket = props.bucket;
+
+    if (props.role) {
+      this.role = props.role;
+    } else {
+      this.role = (this.bucket.node.tryFindChild("ExecutionRole") ??
+        new Role(this.bucket, "ExecutionRole", {
+          assumedBy: new CompositePrincipal(),
+        })) as IRole;
+    }
```
cogwirrel: I think better to create the role by default rather than attempting to reuse one already from the bucket construct - if users are worried about creating too many roles they can pass one in :)

Are you trying to "lazily" allow apigw to assume the role only for a specific method and path? If so I guess the empty `CompositePrincipal` probably works, will just need to update the `grant` to a `grantAssumeRole` I think :)

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: Sorry, I'm not a TS expert 🙈

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: I just did! I like the idea of specifying the actions based on whether it's a read/write/delete :)

Haha I think you scoped down it pretty well! 😄 Just important to swap the path parameters with `*` so IAM allows access for arbitrary values there :) eg. `/pet/{petId}/details.json` becomes `/pet/*/details.json` which means at runtime someone could hit say a `/pet/scoobydoo` API and retrieve the file at `/pet/scoobydoo/details.json` :)

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: So you prefer this more complex approach ? I can keep it?

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: Already done in my last commit

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: But if two or more `S3Integration` are declared for the same bucket, we will have a conflict with roles right?

And yes in order to scope as down as possible I was thinking of declaring an empty role at init (just to store a reference and to get a roleArn in render) and then properly declare a `ServicePrincipal` grant per resource

But maybe it's over-engineering, I don't know, this is why I asked for feedback on this one

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: Absolutely!

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: So I think we want arbitrary path parameters to be mapped as defined in the Smithy/OpenAPI model :) Just mapping them to the same name like `{bucket}` to `{bucket}` is simplest rather than renaming like in the tutorial :)

So I think as long as we add the above mapping it'd map them? Worth testing though! :)

v41e: Ok your're right, I think that the tutorial overcomplicated the thing with different names too

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: Yeah I think let's keep it :)

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: Ahh I'd missed that - that's a very good point! :) I think the thing I found strange was that we're looking for the role in the bucket scope, maybe the simpler approach would be to move everything into the `render` method?

ie the render method could:

* Find or create the role on `props.scope` which is the same scope as the API
* Just create it with the apigw service principal (and if you want scope it down to the api but don't worry about scoping the source arn to the method/path as I think we might fill up the max trust policy size fairly quickly when reusing the role)
* Grant the permissions to the role based on the method (we can't do this in the `grant` method any more if the role is created in the `render` method)
* Then do the rest of what it already does :)

What do you think?

v41e: Looks great!

v41e: It's pushed :)

I'll improve tests, docs and examples later.

Where should I put examples ? I didn't find any with custom grant for lambda

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,162 @@
...
+    this.bucket = props.bucket;
+    this.method = props.method;
+    this.path = props.path;
+    this.errorIntegrationResponse = props.errorIntegrationResponse;
+  }
+
+  private executionRole(scope: IConstruct): IRole {
+    // Retrieve or create the shared S3 execution role
+    return (scope.node.tryFindChild(this.executionRoleId) ??
+      new Role(this.bucket, this.executionRoleId, {
```
cogwirrel: Ooh we should create the role on the passed in `scope` instead of the bucket :)

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: Awesome! Your latest changes look great! Thanks for all your hard work on this! 😄

So the `grant` method just gets called internally by the `TypeSafeRestApi` construct - the idea is the method is just a "hook" for the integration to grant any permissions it needs :)

I mentioned above that I thought we couldn't grant the permissions in the `grant` method any more... but what you've done is actually way cleaner, as it doesn't matter if `render` or `grant` is called first - it just uses the role if it's there or creates one - so great stuff there! 😄

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: 🤦

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: So you probably don't need to talk about any of the grant stuff in the docs :)

packages/type-safe-api/src/construct/integrations/s3.ts
v41e: dumb mode activated

packages/type-safe-api/docs/developer_guides/type-safe-api/integrations.md
```diff
@@ -47,6 +47,31 @@ When you instantiate your generated `Api` construct, you will need to provide an
...
+    // A role to grant access to the S3 bucket to API Gateway, a role with required access is created by default
+    role: yourS3BucketRole
+
+    // The method to use to interact with the S3 bucket, default to integration method
+    method: 'get'
+
+    // The path to use to interact with the S3 bucket, default to integration path
+    path: 'example.json'
+}
+```
```
cogwirrel: Ooh just a thought for the docs, it'd be great to structure it in 2 sections, a code example in all 3 supported languages for the easy default case, then a code example for all 3 for a more complex custom case.

Here's something you can copy/paste and customise! :) I want to show off this cool feature in the docs haha!

````md
## S3 integration

For integrating an API operation with an S3 bucket, use `Integrations.s3`.

=== "TS"

    ```ts
    Integrations.s3({ bucket: yourS3Bucket });
    ```

=== "Java"

    ```java
    Integrations.s3(S3IntegrationProps.builder()
            .bucket(yourS3Bucket)
            .build());
    ```

=== "Python"

    ```python
    Integrations.s3(bucket=your_s3_bucket)
    ```

!!!note
    By default, this will request an S3 object with the HTTP method and path defined in your API, for example:`GET /pets/{petId}` would perform a `GetObject` for `s3://<yourS3Bucket>/pets/{petId}`, where the path parameter `{petId}` is substituted by the user supplied value.

For more information, for example how the HTTP methods map to S3 operations, refer to the [Tutorial: Create a REST API as an Amazon S3 proxy in API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/integrating-api-with-aws-services-s3.html).

### Customise S3 Integrations

You can customise the integration to access objects in S3 at a different key to the API operation path. You can also reference path parameters in the overridden path.

For example, you could have an API operation `POST /pets/{petId}/delete`, but map it to an S3 `DeleteObject` operation for a different object key:

=== "TS"

    ```ts
    Integrations.s3({
        bucket: yourS3Bucket,
        method: "delete",
        path: "my-pets/{petId}/details.json"
    });
    ```

=== "Java"

    ```java
    Integrations.s3(S3IntegrationProps.builder()
            .bucket(yourS3Bucket)
            .method("delete")
            .path("my-pets/{petId}/details.json")
            .build());
    ```

=== "Python"

    ```python
    Integrations.s3(
        bucket=your_s3_bucket,
        method="delete",
        path="my-pets/{petId}/details.json"
    )
    ```

````

Comment (cogwirrel) [reply packages/type-safe-api/docs/developer_guides/type-safe-api/integrations.md]: Actually it's also worth noting about customising the error responses too! :)

File: packages/type-safe-api/src/construct/integrations/error-integration-response.ts [hunk 19]
```diff
@@ -0,0 +1,95 @@
...
+  public abstract render(props: IntegrationRenderProps): {
+    [responseStatusPattern: string]: ApiGatewayIntegrationResponse;
+  };
+}
+
+/**
+ * An error integration response that returns no error responses
+ */
+export class NoneErrorIntegrationResponse extends ErrorIntegrationResponse {
+  public render(): {
```
cogwirrel: Looks like the JSII build failed with the error:
```
error JSII5005: "@aws/type-safe-api.NoneErrorIntegrationResponse#render" has 0 parameters when overriding @aws/type-safe-api.ErrorIntegrationResponse. It should accept 1 parameters
```

I think you might just need to do:

```ts
public render(_props: IntegrationRenderProps): {
```

😄

packages/type-safe-api/docs/developer_guides/type-safe-api/integrations.md
cogwirrel:
````md
### Error Behaviour

By default, if the S3 operation throws an error (for example an object at the requested path does not exist), these will be mapped to an empty `500` error response.

You can customise the error behaviour by passing an `errorIntegrationResponse`.

* `ErrorIntegrationResponses.catchAll()` - default behaviour, returns a `500` response for any S3 error
* `ErrorIntegrationResponses.none()` - TODO @valebedu what happens when S3 errors with this? haha
* `ErrorIntegrationResponses.custom(...)` - custom error responses, mapping an S3 HTTP status regex to the corresponding response API gateway should return
````

File: packages/type-safe-api/src/construct/integrations/error-integration-response.ts (hunk 19 above)
Comment (v41e) [reply packages/type-safe-api/src/construct/integrations/error-integration-response.ts]: You're right I didn't understand where this error came from, thanks for the fix!

Now, tests run but I got the following error:

```
@aws/type-safe-api:   ● Test suite failed to run
@aws/type-safe-api:     Cannot find module 'aws-cdk-lib/aws-s3/lib/perms' from 'src/construct/integrations/s3.ts'
@aws/type-safe-api:     Require stack:
@aws/type-safe-api:       src/construct/integrations/s3.ts
@aws/type-safe-api:       src/construct/integrations/integrations.ts
@aws/type-safe-api:       src/construct/integrations/index.ts
@aws/type-safe-api:       src/construct/index.ts
@aws/type-safe-api:       src/index.ts
@aws/type-safe-api:       test/project/model/type-safe-api-model-project.test.ts
@aws/type-safe-api:       4 | import { Grant, IRole, Role, ServicePrincipal } from "aws-cdk-lib/aws-iam";
@aws/type-safe-api:       5 | import { IBucket } from "aws-cdk-lib/aws-s3";
@aws/type-safe-api:     > 6 | import * as perms from "aws-cdk-lib/aws-s3/lib/perms";
@aws/type-safe-api:         | ^
@aws/type-safe-api:       7 | import * as cxapi from "aws-cdk-lib/cx-api";
@aws/type-safe-api:       8 | import { IConstruct } from "constructs";
@aws/type-safe-api:       9 | import { ErrorIntegrationResponse } from "./error-integration-response";
@aws/type-safe-api:       at Resolver._throwModNotFoundError (../../node_modules/.pnpm/jest-resolve@29.6.4/node_modules/jest-resolve/build/resolver.js:427:11)
@aws/type-safe-api:       at Object.require (src/construct/integrations/s3.ts:6:1)
@aws/type-safe-api:       at Object.require (src/construct/integrations/integrations.ts:6:1)
@aws/type-safe-api:       at Object.require (src/construct/integrations/index.ts:4:1)
@aws/type-safe-api:       at Object.require (src/construct/index.ts:6:1)
@aws/type-safe-api:       at Object.require (src/index.ts:9:1)
@aws/type-safe-api:       at Object.<anonymous> (test/project/model/type-safe-api-model-project.test.ts:4:1)
```

Maybe we're not supposed to used perms directly?

Comment (cogwirrel) [reply packages/type-safe-api/src/construct/integrations/error-integration-response.ts]: Ahh sad times! Yeah sounds like we can't use them directly :(

Perhaps we go with the simpler `grantRead` and `grantReadWrite` option? Alternatively we could copy/paste the perms but that might get a bit annoying to maintain!

Comment (cogwirrel) [reply packages/type-safe-api/src/construct/integrations/error-integration-response.ts]: I suppose they won't change very often so copy/pasting the perms is fine too - happy with whichever you prefer! :)

File: packages/type-safe-api/test/construct/type-safe-rest-api.test.ts (hunk 9 above)
Comment (v41e) [reply packages/type-safe-api/test/construct/type-safe-rest-api.test.ts:346]: Do you know why when running `pnpm run build` I got:

```log
ws/type-safe-api:   ● Generate Mock Data Unit Tests › Generates Mock Data For Recursive Definitions With Required Recursive Reference
@aws/type-safe-api:     Command failed: ../../../../scripts/type-safe-api/custom/mock-data/generate-mock-data --spec-path ../../../resources/specs/recursive-required.yaml --output-path ../../../../../../../../../../../../var/folders/cc/8rdl574s5x90tw1ckq2_zx6m0000gn/T/tmp.wL64GP
@aws/type-safe-api:     /Users/vale/.pdk/type-safe-api/pnpm/node_modules/.pnpm/ts-node@10.9.1_@types+node@20.1.5_typescript@5.0.4/node_modules/ts-node/src/index.ts:859
@aws/type-safe-api:         return new TSError(diagnosticText, diagnosticCodes, diagnostics);
@aws/type-safe-api:                ^
@aws/type-safe-api:     TSError: ⨯ Unable to compile TypeScript:
@aws/type-safe-api:     generate-mock-data.ts(4,34): error TS2307: Cannot find module '@faker-js/faker' or its corresponding type declarations.
@aws/type-safe-api:     generate-mock-data.ts(9,22): error TS2307: Cannot find module 'reregexp' or its corresponding type declarations.
@aws/type-safe-api: 
@aws/type-safe-api:       39 |         const outputPath = path.relative(path.resolve(__dirname), tmpDir);
@aws/type-safe-api:       40 |         const command = `../../../../scripts/type-safe-api/custom/mock-data/generate-mock-data --spec-path ${specPath} --output-path ${outputPath}`;
@aws/type-safe-api:     > 41 |         exec(command, {
@aws/type-safe-api:          |             ^
@aws/type-safe-api:       42 |           cwd: path.resolve(__dirname),
@aws/type-safe-api:       43 |         });
@aws/type-safe-api:       44 |       })
```

v41e:
When viewing your example in docs I was thinking about the fixed status code with a DELETE and 200 is sometimes not what user wants.

So I added a prop to override it if needed

packages/type-safe-api/test/construct/type-safe-rest-api.test.ts
cogwirrel: Ooh you might need to delete the directory we cache script dependencies in (`rm -rf ~/.pdk`), then try rebuilding :)

cogwirrel:
> So I added a prop to override it if needed

Thanks for this! 😄  It made me start thinking that users may even wish to customise other parts of the default/successful response, so I decided to generify your `ErrorIntegrationResponse`- Hope you don't mind! 😄  What do you think?

I was also playing around with a test project using your branch - it's so awesome! Love how path parameters just work and I can change the path/method easily :)

One slightly less intuitive thing I found was the default `catchAll` behaviour returning S3's xml errors in the response body, as the default model we vend for Type Safe API includes 400/403/404/500 error responses as a simple json object with a `message` property:

https://github.com/aws/aws-pdk/blob/09335a2af5607a299608604b2870a076bb9fa55a/packages/type-safe-api/src/project/model/smithy/smithy-definition.ts#L169-L176

Some kind of automatic coercion of the s3 xml error into the corresponding error response in the model would be the ideal way to handle it but I'd have thought that would be super complicated to build! As a compromise, I decided to make the default error behaviour `IntegrationResponseSet.s3JsonErrorMessage()` which extracts the S3 error message and returns it in a way which conforms to the default vended error models :)

cogwirrel (approved):
Awesome work @valebedu! Thanks so much for your contribution!

Given I've added a little bit to this I might get a quick second pair of eyes on this from @agdimech :)

v41e:
Awesome! Thank you so much :)

Yes, once you try S3 you find it really useful, I use it mostly in combinations with cron lambdas that run each several hours to get a specific result from a DB, writing it to a bucket an serving it through API Gateway, this is much cheaper and offers less pressure over DB for a result that doesn't change often.

And yes handling all S3 errors could be quite difficult to do. Maybe if this integration became popular with developers we will improve it :)

packages/type-safe-api/src/construct/integrations/integration-response-set.ts
```diff
@@ -0,0 +1,187 @@
...
+import {
+  ApiGatewayIntegrationResponse,
+  IntegrationRenderProps,
+} from "./integration";
+import { generateCorsResponseParameters } from "../prepare-spec-event-handler/prepare-spec";
+
+/**
+ * A set of integration responses. You can extend this to implement your own integration response set
+ */
+export abstract class IntegrationResponseSet {
```
agdimech, line 12: Given this doesn't implement anything, wouldn't it be better to model this as an interface?

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    this.bucket = props.bucket;
+    this.method = props.method;
+    this.path = props.path;
+    this.integrationResponseSet = props.integrationResponseSet;
+  }
+
+  private executionRole(scope: IConstruct): IRole {
+    // Retrieve or create the shared S3 execution role
+    return (scope.node.tryFindChild(this.executionRoleId) ??
```
agdimech: probably should asser that `scope.node.tryFindChild(this.executionRoleId)` is a Role.

packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -0,0 +1,126 @@
...
+   */
+  public grant({ scope, method, path }: IntegrationGrantProps) {
+    const executionRole = this.executionRole(scope);
+
+    // Replace path parameters with * to grant access to arbitrary values for path parameters
+    const permissionPath = (this.path ?? path).replace(/{[^\}]*\}/g, "*");
+
+    // Grant read access for GET/HEAD/OPTIONS/TRACE, otherwise write
+    if (
+      ["get", "head", "options", "trace"].includes(
```
agdimech, line 129: Will this work generically if a user decides to map a get route to a S3 post?

agdimech (commented):
This is looking really good, nice team work :)

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: Yep, since we use `this.method` by default which is the S3 integration method, and fall back to the API operation method if not defined :)

packages/type-safe-api/src/construct/integrations/integration-response-set.ts
cogwirrel: Probably should be an interface, though I quite like that this is consistent with the `Integration` abstract class (which is an abstract class because it has a default implementation of a `grant` method), and allows us to implement additional functionality into the class in the future without a breaking change. Though I don't foresee a need to add additional functionality any time soon...

packages/type-safe-api/src/construct/integrations/s3.ts
cogwirrel: Yeah good call! :)

agdimech (approved):
Looks good!!

PR #596, opened 2023-10-13
fix(type-safe-api): mock data generation for recursive references

### LINKED ISSUE #545: [BUG] (type-safe-api) mock data generation stackoverflows for recursive structures
### Describe the bug

Mock data generation fails with a stackoverflow for recursive data structures, eg:

```smithy
structure Node {
  left: Node
  right: Node
}
```

### Expected Behavior

Mock data generation to succeed!

### Current Behavior

Fails with a stackoverflow

### Reproduction Steps

Use recursive structure in your smithy/openapi model and generate infrastructure

### Possible Solution

Move away from library for generating mock responses and build our own, or investigate other libraries.

Current workaround is to disable mock generation, eg:

```ts
infrastructure: {
  language: Language.TYPESCRIPT,
  options: {
    typescript: {
      mockDataOptions: {
        disable: true,
      }
    }
  }
},
```

### PDK version used

0.19.67

Typescript, Java, Python

osx

cogwirrel:
The mock data generation library we were using previously stackoverflowed when recursive references were used (which is perfectly valid in openapi). We replace this library with our own implementation, which additionally better conforms to the spec by taking into account min/max/pattern etc constraints.

Fixes #545

### DIFF
--- a/packages/type-safe-api/scripts/type-safe-api/custom/mock-data/generate-mock-data
+++ b/packages/type-safe-api/scripts/type-safe-api/custom/mock-data/generate-mock-data
@@ -7,11 +7,13 @@ spec_path=''
 output_path=''
 locale='en'
 max_array_length='3'
+seed='1337'
 while [[ "$#" -gt 0 ]]; do case $1 in
   --spec-path) spec_path="$2"; shift;;
   --output-path) output_path="$2"; shift;;
   --locale) locale="$2"; shift;;
   --max-array-length) max_array_length="$2"; shift;;
+  --seed) seed="$2"; shift;;
 esac; shift; done
 
 echo "Generating Mock Data..."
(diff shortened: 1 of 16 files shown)


packages/type-safe-api/scripts/type-safe-api/custom/mock-data/generate-mock-data
```diff
@@ -7,11 +7,13 @@ spec_path=''
 output_path=''
 locale='en'
 max_array_length='3'
+seed='1337'
```
agdimech, line 10: haha love it

agdimech (approved):

packages/type-safe-api/scripts/type-safe-api/custom/mock-data/generate-mock-data
cogwirrel: Couldn't resist! 😂

PR #597, opened 2023-10-14
feat(type-safe-api): enable snapstart for java handlers by default and optimise java interceptors

### LINKED ISSUE #572: [FEATURE] (type-safe-api) Enable snapstart for java handlers by default
### Describe the feature

It would be handy to have [SnapStart](https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html) enabled on the generated handler function constructs for java, given it reduces the cold start time significantly.

### Use Case

Speed!

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### PDK version used

0.22.1

Java

MacOS

cogwirrel:
Java handlers now have snapstart enabled by default to improve cold start times.

Ensure logging, metrics and tracing interceptors perform as much initialisation as possible during the init phase which can therefore be saved from snapstart invocations.

Also move the annotation processing to the init phase since this takes approximately a second.

This reduces cold start times from ~25s down to ~4s. Warm invocations remain the same at <100ms.

On a cold start it seems as though the logger takes about 500ms per log, and the tracer takes about 1500ms to end the subsegment - so there may be an opportunity to reduce the 4s cold start even further if we can figure out why :)

Fixes #572

### DIFF
--- a/packages/type-safe-api/scripts/type-safe-api/generators/java-cdk-infrastructure/templates/functions.handlebars
+++ b/packages/type-safe-api/scripts/type-safe-api/generators/java-cdk-infrastructure/templates/functions.handlebars
@@ -15,12 +15,13 @@
 
 import org.jetbrains.annotations.NotNull;
 import software.amazon.awscdk.services.lambda.Function;
+import software.aws.pdk.type_safe_api.SnapStartFunction;
 import software.constructs.Construct;
 
 /**
  * Lambda function construct which points to the {{vendorExtensions.x-handler.language}} implementation of {{operationIdCamelCase}}
  */
-public class {{operationIdCamelCase}}Function extends Function {
+public class {{operationIdCamelCase}}Function extends {{#startsWith vendorExtensions.x-handler.language 'java' ~}}SnapStart{{~/startsWith}}Function {
     public {{operationIdCamelCase}}Function(@NotNull Construct scope, @NotNull String id, @NotNull {{operationIdCamelCase}}FunctionProps props) {
         super(scope, id, props);
     }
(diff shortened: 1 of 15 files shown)


packages/type-safe-api/scripts/type-safe-api/generators/python-cdk-infrastructure/templates/functions.handlebars
```diff
@@ -10,6 +10,7 @@
 from aws_cdk.aws_lambda import (
     Function, Runtime, Tracing, Code
 )
+from aws_pdk.type_safe_api import SnapStartFunction
```
agdimech, line 13: Shouldn't this be conditional?

packages/type-safe-api/scripts/type-safe-api/generators/java-cdk-infrastructure/templates/functions.handlebars
```diff
@@ -67,6 +68,7 @@ import software.amazon.awscdk.services.lambda.VersionOptions;
 import software.amazon.awscdk.services.logs.RetentionDays;
 import software.amazon.awscdk.services.sns.ITopic;
 import software.amazon.awscdk.services.sqs.IQueue;
+import software.aws.pdk.type_safe_api.SnapStartFunctionProps;
```
agdimech, line 71: Shouldn't this be conditional (same as line 56)

packages/type-safe-api/scripts/type-safe-api/generators/python-cdk-infrastructure/templates/functions.handlebars
cogwirrel: The generated infra code doesn't error for unused imports so I kept them there for simplicity - otherwise I need to do both "if python" and "if typescript" as there's no "else" - doable but not much gain besides slightly neater generated code :)

agdimech (approved):

PR #610, opened 2023-10-20
fix(type-safe-api): fix syntax error for python runtime request parameters with descriptions

### LINKED ISSUE #609: [BUG] (type-safe-api) Python runtime has invalid code when request parameters have descriptions
### Describe the bug

When request parameters have descriptions, eg:

```smithy
@readonly
@http(method: "GET", uri: "/hello")
@handler(language: "python")
operation SayHello {
    input := {
        /// This is a description
        @httpQuery("name")
        @required
        name: String
    }
    output := {
        @required
        message: String
    }
    errors: [NotFoundError]
}
```

The generated python code is invalid:

```python

class SayHelloRequestParameters(BaseModel):
    """
    Query, path and header parameters for the SayHello operation
    """
    name: Annotated[StrictStr, Field(..., description&#x3D;&quot;This is a description&quot;)] = Field(alias='name')
```

### Expected Behavior

Working code when parameters have descriptions!

### Current Behavior

Invalid code!

### Reproduction Steps

Add description to a request parameter (path/query/header).

### Possible Solution

_No response_

_No response_

### PDK version used

0.22.17

Python

OSX

cogwirrel:
Descriptions in Python request parameters would previously render invalid python syntax as quotes were escaped as `&quot;`. Use triple curly braces (`{{{vendorExtensions.x-py-typing}}}`) to ensure these are rendered correctly.

Additionally, field references can only be used once per field in a pydantic model, and since fields with descriptions include the field in the type annotation these would fail at runtime when another field is used in the assignment, ie:

```python
foo: Annotated[StrictStr, Field(..., description="...")] = Field(alias='foo')
```

The above causes a runtime error (```ValueError: `Field` default cannot be set in `Annotated` for ...``` - https://github.com/pydantic/pydantic/issues/3991 )

We therefore remove the assignment of a `Field` value, which is ok since the alias is not required as the handler wrapper already constructs a dict with the `paramName`s as keys rather than the `baseName`s which are in the raw request. ie we already do:

```mustache
"{{paramName}}": coerce_parameter("{{baseName}}", ...)
```

Which means we even load renamed parameters (like a header `x-foo-bar` from a dictionary which names it as `x_foo_bar`).

Fixes #609

### DIFF
--- a/packages/type-safe-api/scripts/type-safe-api/generators/python/templates/operationConfig.mustache
+++ b/packages/type-safe-api/scripts/type-safe-api/generators/python/templates/operationConfig.mustache
@@ -1,7 +1,7 @@
 from __future__ import annotations
 import urllib.parse
 import json
-from typing import Callable, Any, Dict, List, NamedTuple, TypeVar, Generic, Union, TypedDict, Protocol, Optional, Literal
+from typing import Callable, Any, Dict, List, NamedTuple, TypeVar, Generic, Union, TypedDict, Protocol, Optional, Literal, Annotated
 from functools import wraps
 from dataclasses import dataclass, fields
 from datetime import datetime
@@ -215,7 +215,7 @@ class {{operationIdCamelCase}}RequestParameters(BaseModel):
     """
     {{#allParams}}
     {{^isBodyParam}}
-    {{paramName}}: {{vendorExtensions.x-py-typing}} = Field(alias='{{baseName}}'{{^required}}, default=None{{/required}})
+    {{paramName}}: {{{vendorExtensions.x-py-typing}}}
     {{/isBodyParam}}
     {{/allParams}}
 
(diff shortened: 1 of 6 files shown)


packages/type-safe-api/scripts/type-safe-api/generators/python/templates/operationConfig.mustache
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 import urllib.parse
 import json
-from typing import Callable, Any, Dict, List, NamedTuple, TypeVar, Generic, Union, TypedDict, Protocol, Optional, Literal
+from typing import Callable, Any, Dict, List, NamedTuple, TypeVar, Generic, Union, TypedDict, Protocol, Optional, Literal, Annotated
```
agdimech, line 4: How are we using `Annotated`?

agdimech (approved):

packages/type-safe-api/scripts/type-safe-api/generators/python/templates/operationConfig.mustache
cogwirrel: The fields with descriptions use it, ie `{{paramName}}: {{{vendorExtensions.x-py-typing}}}` might become `name: Annotated[StrictStr, Field(..., description="This is a description")]` 😄

PR #614, opened 2023-10-25
fix: prevent error by running pnpm

drskur:
When I created infrastructure project using pnpm, the pipeline ran pnpm and the following error occurred.

```
|[Container] 2023/10/25 07:42:20.717048 Running command npx projen install|
|25|npm WARN exec The following package was not found and will be installed: projen@0.76.4|
|26|/bin/sh: 1: pnpm: not found|
|27|👾 unable to evaluate environment variable PATH=$(pnpm -c exec "node --print process.env.PATH"): unknown error|
|28||
|29|[Container] 2023/10/25 07:42:27.864655 Command did not exit successfully npx projen install exit status 1|
|30|[Container] 2023/10/25 07:42:27.867954 Phase complete: INSTALL State: FAILED|
|31|[Container] 2023/10/25 07:42:27.867971 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: npx projen install. Reason: exit status 1|
```

This patch makes the Synth process take a little longer, but it can fix the problem easily and reliably.

Fixes #


packages/pipeline/src/pdk-pipeline.ts
```diff
@@ -315,7 +315,7 @@ export class PDKPipeline extends Construct {
               BRANCH: branch,
             }
           : undefined,
-      installCommands: ["npm install -g aws-cdk", "npx projen install"],
+      installCommands: ["npm install -g aws-cdk pnpm", "npx projen install"],
```
cogwirrel, line 318: Thanks for this! Users can override the install/build commands themselves to achieve this result but I think it makes sense to have the default commands work regardless of your package manager (or language).

I think it'd be good to additionally install `@aws/pdk` and then we can replace the `npx projen install` and `npx projen build` commands with `pdk install` and `pdk build` :)

Also, it looks like the build is failing due to the snapshots - please could you build locally and include the updated snapshots in your PR? Thanks!

PR #629, opened 2023-11-01
fix(infrastructure): fix java compilation issue without api, and skip unused constructs in java/py

cogwirrel:
The Java website construct didn't compile when an api wasn't present due to some extra brackets being commented out, and a duplicate constructor with the same signature.

Additionally the website construct would always be rendered for typescript even if no website was configured. Likewise for Java and Python, the api and website constructs were always rendered.

This PR also adds tests for the infrastructure projects.


agdimech (approved):
ahh good catch - not sure how this happend :(

PR #630, opened 2023-11-02
feat(infrastructure): add deploy:dev task to infrastructure projects for cdk hotswap deployment

### LINKED ISSUE #599: [FEATURE] (type-safe-api) Local dev server without SAM
### Describe the feature

Currently, local development requires the use of SAM: https://aws.github.io/aws-pdk/developer_guides/type-safe-api/troubleshooting.html#how-do-i-debug-my-api-locally

There are a few limitations with this approach, namely:

* SAM local API does not support mock integrations (and therefore does not respond to CORS OPTIONS requests)
* SAM local uses containers to run lambda functions, which while is more realistic, slows the iteration cycle (particularly on an M1 Mac)
* Need to specify the `type-safe-api-local=true` CDK context variable to synth something that SAM local can run but remember not to check this in or deploy with it set

It would be great to have a more integrated local dev experience to help for faster iteration.

### Use Case

Usually testing out the API locally has no real dev cycle improvements vs just waiting for a deployment, given the docker image startup and response time. And pointing a website at a local server isn't very helpful right now since the CORS OPTIONS requests return 500s.

It would be great to have a seamless way to start a local server (whether implementing handlers in TS, Java or Python or a combination of all 3).

### Proposed Solution

Possibly having a "frontend" server which proxies requests to another server per language such that the multi-language handlers use case is covered?

Each language specific server would need to implement request/response mappers similar to https://aws.github.io/aws-pdk/developer_guides/type-safe-api/custom_integration_ecs.html#mapping-requests-and-responses

### Other Information

Things to consider:

* It should natively support mock integrations and lambda integrations
* It should ideally "live reload", ie if you save changes to a handler function the next request should be served by that handler code
* How are AWS credentials are managed? Do we just expect the environment to have credentials or do we try something smarter, like attempting to assume the role assigned to the lambda function in CDK?
* How do we manage environment variables, often used for things DynamoDB table names or other config?
* How do we manage integrations that aren't to lambda functions, eg S3 or custom integrations? (Likely just return a 500 error to keep things simple to start with)

### Acknowledgements

### PDK version used

0.22.13

Typescript, Java, Python

OSX

cogwirrel:
Adds a `deploy:dev` task to the infrastructure projects which performs a CDK hotswap deployment if possible in order to reduce dev cycle times.

Given the dev cycle time reduction we consider this a better option than a local dev server, since local dev server requires managing credentials locally, as well as things like environment variables for pointing to other resources (eg ddb tables).

Fixes #599

### DIFF
--- a/docs/content/getting_started/shopping_list_app.md
+++ b/docs/content/getting_started/shopping_list_app.md
@@ -708,7 +708,7 @@ We are now ready to deploy our API. To do so, run the following steps:
 ```bash
 pdk build
 cd packages/infra
-pdk run deploy --require-approval never
+pdk deploy:dev
 ```
 
 Once the deployment completes, we can test our API by navigating the the website (either via Cloudfront or locally) and trying out the API Explorer.
@@ -1192,7 +1192,7 @@ If you are happy with your website locally, you can go ahead and deploy it to AW
 ```bash
 pdk build
 cd packages/infra
-pdk run deploy --require-approval never
+pdk deploy:dev
 ```
 
 Once the deployment completes, navigate to your cloudfront URL to play around with your deployed website.
--- a/docs/content/getting_started/your_first_aws_pdk_project.md
+++ b/docs/content/getting_started/your_first_aws_pdk_project.md
@@ -86,7 +86,7 @@ Inspecting the `projenrc` file, we notice that a single construct is instantiate
     - Any parameter listed here can be passed in via the `pdk new` command i.e: `--name="some-other-name"`.
 - For python, the moduleName defaults to a snake-cased project name.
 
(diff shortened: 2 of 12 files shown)


docs/content/getting_started/shopping_list_app.md
```diff
@@ -708,7 +708,7 @@ We are now ready to deploy our API. To do so, run the following steps:
 ```bash
 pdk build
 cd packages/infra
-pdk run deploy --require-approval never
+pdk deploy:dev
```
agdimech, line 711: does this work with pnpm?

cogwirrel: Yep :) `pdk deploy:dev` works but `pdk deploy` doesn't - have updated the script so `pdk deploy` works for pnpm too :)

agdimech (approved):

PR #633, opened 2023-11-03
feat(type-safe-api): java: improve first invoke time and provide warmup hook to prime handlers

cogwirrel:
Further optimisation of the Java handlers, taking the first invocation time down from ~6s to ~1-2s by priming each interceptor's handle method, as well as parts of the handler wrapper.

Additionally provide a `warmUp` method which can be overridden by a Java handler to run code to prime the function, which is called prior to the SnapStart snapshot being taken.


agdimech (approved):
Nice improvement. So much boilerplate for java...

cogwirrel:
Thanks! Haha yeah sooo many lines of code!

PR #640, opened 2023-11-06
feat(type-safe-api): generate handler tests

### LINKED ISSUE #570: [FEATURE] (type-safe-api) Generate tests for handler packages
### Describe the feature

It would be great to generate an example test for each generated handler.

### Use Case

Reduce time to write tests. Demonstrate how to test in terms of modelled types rather than api gateway events.

### Proposed Solution

Generate tests when `@handler` trait is used.

Probably want to generate just once when the handler stub is first generated, but not add it back if the user deletes the test - ie only generate if there is no handler stub present already. (Rather than the handler stub only generating when it doesn't already exist).

### Other Information

_No response_

### Acknowledgements

### PDK version used

0.22.0

Typescript, Java, Python

MacOS

cogwirrel:
* Generate example junit tests for java handlers.
* Generate example pytest tests for python handlers.
* Generate example jest tests for typescript handlers.

Test generation is tied to handler generation, ie if the handler already exists the test won't be generated - it is only generated when a handler is newly generated. This provides a smoother upgrade path as tests won't be generated for existing operations in existing projects, and allows users to delete generated tests without worrying about them reappearing on the next build.

### Additional Changes

Java:
* Adds a lombok `@Builder` to the generated java input classes to make them easier to construct for tests.
* Change the use of the logging interceptor in the generated java handler stubs to create a logger for the class, which is a more standard practice, and also simplifies the test as we don't need to put a logger in the interceptor context.

Fixes #570


packages/type-safe-api/docs/developer_guides/type-safe-api/lambda_handlers.md
```diff
@@ -195,7 +201,7 @@ This will give you generated lambda handler stubs which look like the following:
          */
         @Override
         public SayHelloResponse handle(final SayHelloRequestInput request) {
-            LoggingInterceptor.getLogger(request).info("Start SayHello Operationnn");
+            log.info("Start SayHello Operation");
```
agdimech, line 204: much better :)

agdimech (approved):
Great stuff mate!

PR #645, opened 2023-11-10
fix(monorepo-py): remove incorrect typescript pdk dependency for python monorepos

### LINKED ISSUE #643: [BUG] (cannot create new monorepo-py)
### Describe the bug

When creating a new monorepo in Python via command `pdk new monorepo-py` the installation fails.

### Expected Behavior

Installation finishes successfully with new Python monorepo created.

### Current Behavior

When installing Poetry dependencies, the process fails with:

```
Resolving dependencies... (1.0s)

Because pdk-python1 depends on @aws/pdk (*) which doesn't match any versions, version solving failed.
/Users/falnyr/projects/pdk-python1/node_modules/projen/lib/task-runtime.js:155
                    throw new Error(`Task "${this.fullname}" failed when executing "${command}" (cwd: ${(0, path_1.resolve)(cwd ?? this.workdir)})`);
                    ^

Error: Task "install" failed when executing "poetry update" (cwd: /Users/falnyr/projects/pdk-python1)
```

### Reproduction Steps

Run `pdk new monorepo-py` in an empty folder.

### Possible Solution

Current workaround is to remove `dev_deps=["@aws/pdk"]` dev dependency from `.projenrc.py` and run `pdk` again.

Poetry (version 1.7.0)
Python 3.11.5

### PDK version used

0.22.13

Python

MacOS Sonoma 14.1.1

cogwirrel:
The projen new command adds a dependency on `@aws/pdk` when creating a new python monorepo, which causes the initial install to fail as `@aws/pdk` is not the correct name for the module in pypi. The workaround is for users to remove this dependency, but we can remove it within the monorepo for a smoother initial bootstrap experience for python.

Also adds a basic snapshot test for a python monorepo.

Fixes #643


agdimech (approved):
Nice!

PR #648, opened 2023-11-14
fix(static-website): use node 18 runtime for webacl custom resource lambda for aws sdk v3

cogwirrel:
The upgrade to SDK v3 in the web acl custom resource missed updating the lambda runtime to node 18 which contains the SDK v3.


JeremyJonas (approved):

martenpayne (approved):

cogwirrel:
Ref: https://github.com/aws/aws-pdk/pull/632

PR #660, opened 2023-11-23
feat(static-website): add support for Origin Access Control (OAC)

### LINKED ISSUE #49: [FEATURE] Support zero-config KMS encryption for static-website.
### Describe the feature

Cloudfront OAI does not currently support KMS encryption on buckets and as such a user needs to use S3_MANAGED encryption or follow this guide: https://aws.amazon.com/blogs/networking-and-content-delivery/serving-sse-kms-encrypted-content-from-s3-using-cloudfront/.

### Use Case

Highly regulated customers which mandate using customer managed keys i.e. banks.

### Proposed Solution

A Lambda@Edge function will need to be associated to the distribution which will sign the request on OAI's behalf as per: https://aws.amazon.com/blogs/networking-and-content-delivery/serving-sse-kms-encrypted-content-from-s3-using-cloudfront/.

This lambda must exist in us-east-1 and as such a custom resource will need to be created such that it can be created agnostic of the host region. This lambda will need to intercept origin requests to S3, sign the request using it's own credentials (s3 ro access + kms decrypt).

### Other Information

_No response_

### Acknowledgements

### PDK version used

N/A

Typescript, Java, Python

N/A

agdimech:
This PR exposes a new `OriginAccessControl` construct which is responsible for updating the Distribution and Bucket to configure OAC. As part of this is does the following:

1. Removes any and all references/resources/policies for OAI.
2. Creates an OAC resource
3. Uses a custom resource to update the bucket and optional kms key resource policies which solves an issue related to cyclic dependencies (distro needs bucket but bucket policy needs distro). 
4. Works for SSE and KMS customer keys.

The StaticWebsite construct now also accepts an optional `originAccessType` which by default uses OAI :( but can be configured to use OAC (this is needed to preserve backward compat).

_Note: This implementation is a stop-gap until https://github.com/aws/aws-cdk/issues/21771 is resolved properly. Eventually, the OriginAccessControl component in here will be deprecated in favor of whatever CDK L2 construct(s) come along._

Fixes #49


packages/static-website/src/update_policy_handler/index.ts
```diff
@@ -0,0 +1,114 @@
...
+): Promise<CdkCustomResourceResponse> => {
+  const response: CdkCustomResourceResponse = {
+    StackId: event.StackId,
+    RequestId: event.RequestId,
+    LogicalResourceId: event.LogicalResourceId,
+    PhysicalResourceId: ctx.logGroupName,
+  };
+
+  switch (event.RequestType) {
+    case "Create":
```
agdimech, line 39: Think it's worth adding **Update** to ensure these policies are present in the event a user manually removes them?

**Delete** shouldn't be needed as that will be handled by the delete handler for `BucketPolicy`

packages/static-website/src/origin-access-control.ts
```diff
@@ -0,0 +1,125 @@
...
+export class OriginAccessControl extends Construct {
+  public readonly oac;
+  constructor(scope: Construct, id: string, props: OriginAccessControlProps) {
+    super(scope, id);
+
+    const { distribution, bucket } = props;
+    const stack = Stack.of(this);
+    this.oac = new CfnOriginAccessControl(this, `${id}-OAC`, {
+      originAccessControlConfig: {
+        name: `${stack.stackName}_OAC`,
```
cogwirrel, line 45: If we need to hardcode a name it's best to include something like `this.node.addr.slice(-8)` for some additional uniqueness, eg if we need to add 2 websites in the same stack :)

packages/static-website/src/origin-access-control.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    const distributionArn = stack.formatArn({
+      service: "cloudfront",
+      resource: "distribution",
+      region: "",
+      resourceName: distribution.distributionId,
+      arnFormat: ArnFormat.SLASH_RESOURCE_NAME,
+    });
+    const onEventHandler = new NodejsFunction(this, "UpdatePolicy", {
+      entry: path.join(__dirname, "update_policy_handler/index.ts"),
```
cogwirrel, line 91: Will this still be `.ts` when used as a lib?

packages/static-website/src/update_policy_handler/index.ts
cogwirrel: Yeah probably useful! If doing update you might need to add IDs to the policy statements you add and replace if they exist to avoid continually growing the bucket/key policies :)

cogwirrel (commented):
Looks great! Nice one :)

PR #665, opened 2023-12-02
chore: Replace deprecated command with environment file

jongwooo:
## Description

Update workflows to use environment file instead of deprecated `set-output` command. 
For more information, see: [https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

I found the workflow files that use `set-output` command through the following command:

```bash
$ find . -name '*.yml' -o -name '*.yaml' | xargs egrep '\bset-output\b'
```

**AS-IS**

```yaml
run: echo ::set-output name=latest_commit::"$(git ls-remote origin -h ${{ github.ref }} | cut -f1)"
```

**TO-BE**

```yaml
run: echo "latest_commit=$(git ls-remote origin -h ${{ github.ref }} | cut -f1)" >> $GITHUB_OUTPUT
```

### DIFF
--- a/.github/workflows/aws-arch-pricing.yml
+++ b/.github/workflows/aws-arch-pricing.yml
@@ -35,12 +35,12 @@ jobs:
         # ignore diff if only the pricing manifest json was modified
         run: |-
           git add .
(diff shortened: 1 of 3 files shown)


agdimech (approved):
Thank you such much for your contribution and proactively raising this :)

PR #670, opened 2023-12-14
fix(type-safe-api): use dev+peer deps for generated infrastructure

zsstiers:
Pulling the JSII-compatible deps out to dev+peer improves the ability to use generated infrastructure as a bundled dependency of a construct library. With these dependencies as normal deps they get bundled in and can lead to duplicate, possibly conflicting, versions of these libraries.


cogwirrel (approved):
Thanks for this! Looks great :)

PR #688, opened 2024-01-23
feat(type-safe-api): support configuring lambda handler runtime versions

### LINKED ISSUE #672: [FEATURE] (type-safe-api) Ability to configure runtime/language versions in one place
### Describe the feature

Related to https://github.com/aws/aws-pdk/pull/671 - we need to ensure the Python version in the poetry env is the same as the one used as the lambda runtime.

It would be great to be able to configure this, eg specify Python 3.12 in the .projenrc and have the generated function constructs honour this version, as well as the poetry env.

The same should be true for Node (eg esbuild should target the node version used in the lambda runtime), and Java (eg the pom.xml language version should match the version used in the lambda runtime).

### Use Case

Customise runtime versions

### Proposed Solution

Possibly a configuration option in the .projenrc

### Other Information

_No response_

### Acknowledgements

### PDK version used

0.22.43

Typescript, Java, Python

OSX

cogwirrel:
Users can now specify the runtime version for python, java and typescript (node) lambda handlers. As well as the lambda runtime version, this configures settings for the project itself to target that particular runtime, and is therefore configured in the projenrc.

Fixes #672


JeremyJonas (approved):
Beautiful 🥇

PR #694, opened 2024-01-30
feat: upgrade projen to 0.79

### LINKED ISSUE #661: [BUG] unmet peer projen@^0.76.25: found 0.77.3
### Describe the bug

When initialising a pdk monorepo project (via `pdk new --package-manager=pnpm monorepo-ts`), pnpm warns you that `@aws/pdk` has an unmet peer dependency.

```
 WARN  Issues with peer dependencies found
.
└─┬ @aws/pdk 0.22.42
  └── ✕ unmet peer projen@^0.76.25: found 0.77.3
```

This also occurs if pdk / pnpm decides it needs to install anything (eg. by bumping a version of some other dependency in your `.projenrc.ts` file)

### Expected Behavior

No peer dependency warnings when you set up a new project

### Current Behavior

A peer dependency warning when setting up a new project

### Reproduction Steps

1. using node `v18.18.2`
2. `npm i -g @aws/pdk@latest`
3. `npm i -g pnpm@latest`
4. `mkdir test-pdk && cd test-pdk`
5. `pdk new --package-manager=pnpm monorepo-ts`

During the process you should see the warning outlined above

### Possible Solution

Maybe related to how `npm` handles caret version ranges for 0.x

https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004

(though I am not sure if the same applies to `pnpm`)

_No response_

### PDK version used

0.22.42

Typescript

macos sonoma 14.1

cogwirrel:
Upgrade projen to ^0.79.6

~It may also be worth "pinning" the projen version for the projects PDK exports to the same as the peer dep version to avoid this "unmet peer" issue going forward.~

This PR also sets a default version of projen at the monorepo level, and ensures the projen version is synchronised across all subprojects. This default version is the same as the PDK peer dependency version (since PDK dogfoods the TS monorepo), which ensures we won't hit the "unmet peer" issue when projen bump to 0.80 and beyond, and we can update projen in our own time.

Note that projen's test behaviour of synthing to a tmp dir when `NODE_ENV=test` was set has changed, so I needed to set a different environment variable to avoid breaking all our tests :)

Fixes #661


agdimech (approved):

packages/monorepo/src/components/projen-dependency.ts
```diff
@@ -0,0 +1,28 @@
+/*! Copyright [Amazon.com](http://amazon.com/), Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0 */
+import { DependencyType, Project } from "projen";
+
+export const DEFAULT_PROJEN_VERSION = "^0.79.6";
```
agdimech: Having this defined like this may be a little difficult to maintain. What if we did something like:

```
import projenVersion from './projen-version.txt';
export const DEFAULT_PROJEN_VERSION = projenVersion;
```

Then in `projen-version.txt` we have simply:

```
^0.79.6
```

This will allow us to set up a cron workflow weekly to pull down the latest version, update the `projen-version.txt` file with no user intervention (besides a PR approval).

agdimech (commented):
Looks good - shall we include the workflow to auto update this as well?

cogwirrel:
> Looks good - shall we include the workflow to auto update this as well?

I'll send a followup PR :)

PR #695, opened 2024-02-01
build: update projen version as part of upgrade workflow

cogwirrel:
Adds a task to run as part of the `upgrade-mainline` workflow which updates the projen version such that PDK and its vended projects all depend on the same projen version.


agdimech (approved):
Looks good / thanks!

PR #703, opened 2024-02-20
refactor: remove sharp in favour of pure ts-graphviz

agdimech:
This PR removes the need for sharp in favour of using pure ts-graphviz. This results in a significant (500%) reduction in the overall pdk distributable size which will result in quicker installs from a consumer side and faster packaging from a authoring side :)


agdimech:
Build issue is caused by a required global dependency on [graphviz](https://github.com/ts-graphviz/ts-graphviz/blob/main/src/adapter/utils/create-command-and-args.ts).

It is also documented here: https://github.com/ts-graphviz/ts-graphviz?tab=readme-ov-file#ts-graphvizadapter-module-

I am going to have to add a check to ensure dot is available globally.

Two options:

**Option A:** Fail the build if `dot` is not found.
**Option B:** Log a warning saying that PNG and SVG is not supported until `graphviz` is installed globally and as such generation will be skipped. This will mean dot will be the only format supported until graphviz is installed globally.

_My recommendation is **Option B** as to not break existing consumers._

JeremyJonas:
> Build issue is caused by a required global dependency on [graphviz](https://github.com/ts-graphviz/ts-graphviz/blob/main/src/adapter/utils/create-command-and-args.ts).
> 
> It is also documented here: https://github.com/ts-graphviz/ts-graphviz?tab=readme-ov-file#ts-graphvizadapter-module-
> 
> I am going to have to add a check to ensure dot is available globally.
> 
> Two options:
> 
> **Option A:** Fail the build if `dot` is not found. **Option B:** Log a warning saying that PNG and SVG is not supported until `graphviz` is installed globally and as such generation will be skipped. This will mean dot will be the only format supported until graphviz is installed globally.
> 
> _My recommendation is **Option B** as to not break existing consumers._

Agree with Option B. The dot file can still be used downstream outside of the build, the PNG and SVG are extra features for auto gen in my mind - so  100% aligned on warning.

packages/cdk-graph-plugin-diagram/src/internal/graphviz/entities/nodes.ts
```diff
@@ -109,25 +107,15 @@ export class ImageNode extends Node {
...
-
-      this.attributes.set(
-        "height",
-        baseHeight + labelLines * LABEL_LINE_HEIGHT
-      );
     } else {
       this.attributes.set("labelloc", "c");
-      this.attributes.set("height", baseHeight);
+      this.attributes.set("penwidth", 0.25);
     }
```
JeremyJonas: Remember with our dynamic height based on "label lines height" the rendering was jank. If I remember correctly there was overlap and incorrect edge placement. Did you get that resolved or is no longer an issue?

packages/cdk-graph-plugin-diagram/src/internal/graphviz/entities/nodes.ts
```diff
@@ -143,11 +131,7 @@ export class CfnResourceNode extends ImageNode {
 
     this.attributes.apply(GraphTheme.instance.cfnResourceNode);
 
-    this.resize(
-      GraphTheme.instance.cfnResourceNode.height === ""
-        ? undefined
-        : GraphTheme.instance.cfnResourceNode.height
-    );
+    this.resize();
```
JeremyJonas: These resize inputs enable the theme to override the defaults, are you punting this or have it handled somewhere else? Either is fine for now, given we don't have much in terms of theming atm.

packages/cdk-graph-plugin-diagram/src/internal/graphviz/entities/nodes.ts
JeremyJonas: Looks like the image based node labeling is "right aligned" (instead of centered), and the edge connecting to image nodes is not centered as well - so to my above comment, this code was to correct those alignment issues.

image

packages/cdk-graph-plugin-diagram/src/internal/utils/svg.ts
```diff
@@ -3,121 +3,21 @@ SPDX-License-Identifier: Apache-2.0 */
...
-      {
-        name: "",
-        type: "text",
-        attributes: {},
-        children: [],
-        value: FONT_CSS_CLASSES,
-      },
-    ],
-  });
-}
```
JeremyJonas: Does graphviz to svg output preserve the font classes?

projenrc/projects/cdk-graph-plugin-diagram-project.ts
```diff
@@ -34,24 +33,20 @@ export class CdkGraphPluginDiagramProject extends CdkGraphPluginProject {
         "@types/sharp",
```
JeremyJonas: Can remove `@types/sharp` as well

projenrc/projects/cdk-graph-plugin-diagram-project.ts
```diff
@@ -34,24 +33,20 @@ export class CdkGraphPluginDiagramProject extends CdkGraphPluginProject {
...
       bundledDeps: [
-        "@hpcc-js/wasm",
         "execa@5.1.1",
         "fs-extra",
         "he",
         "lodash.clonedeep",
         "lodash.startcase",
         "lodash.uniqby",
         "lodash.words",
-        "sharp",
```
JeremyJonas: Love seeing these go away - nice work mate!!

JeremyJonas (commented):
Awesome work mate, great reduction in complexity and size!

Looks like I missed the adapter support in ts-graphviz by a month (https://github.com/ts-graphviz/ts-graphviz/releases/tag/v1.3.0) 😆 😆 

Only concern the image node rendering of labels and edges, alignment is off as called out in comments.

projenrc/projects/cdk-graph-plugin-diagram-project.ts
agdimech: :)

projenrc/projects/cdk-graph-plugin-diagram-project.ts
agdimech: Ahh yep - will do!

packages/cdk-graph-plugin-diagram/src/internal/utils/svg.ts
agdimech: Ahh forgot to add this back, will update.

packages/cdk-graph-plugin-diagram/src/internal/graphviz/entities/nodes.ts
agdimech: Added them back in.

JeremyJonas (approved):

Issue #508, opened 2023-07-19
[BUG] (static-website): WAF creation custom resource returns WAFUnavailableEntityException


jessieweiyi:
### Describe the bug

My PDK app includes PDK static-website construct and the static-website construct has the cidrAllowList configured. 

When I deployed the app to the us-east-1 region, the deployment failed for most of times (4 out of 5 times) because the WAF creation custom resource returns WAFUnavailableEntityException. 

Below is the CloudWatch log for the WAF custom resource lambda:

```
{
    "errorType": "WAFUnavailableEntityException",
    "errorMessage": "AWS WAF couldn't retrieve the resource that you requested. Retry your request.",
    "code": "WAFUnavailableEntityException",
    "message": "AWS WAF couldn't retrieve the resource that you requested. Retry your request.",
    "time": "2023-07-17T06:45:23.046Z",
    "requestId": "f2e343bb-86e3-441f-99b6-59a88fc139e4",
    "statusCode": 400,
    "retryable": false,
    "retryDelay": 16.044677277845267,
    "stack": [
        "WAFUnavailableEntityException: AWS WAF couldn't retrieve the resource that you requested. Retry your request.",
        "    at Request.extractError (/var/runtime/node_modules/aws-sdk/lib/protocol/json.js:61:27)",
        "    at Request.callListeners (/var/runtime/node_modules/aws-sdk/lib/sequential_executor.js:106:20)",
        "    at Request.emit (/var/runtime/node_modules/aws-sdk/lib/sequential_executor.js:78:10)",
        "    at Request.emit (/var/runtime/node_modules/aws-sdk/lib/request.js:686:14)",
        "    at Request.transition (/var/runtime/node_modules/aws-sdk/lib/request.js:22:10)",
        "    at AcceptorStateMachine.runTo (/var/runtime/node_modules/aws-sdk/lib/state_machine.js:14:12)",
        "    at /var/runtime/node_modules/aws-sdk/lib/state_machine.js:26:10",
        "    at Request.<anonymous> (/var/runtime/node_modules/aws-sdk/lib/request.js:38:9)",
        "    at Request.<anonymous> (/var/runtime/node_modules/aws-sdk/lib/request.js:688:12)",
        "    at Request.callListeners (/var/runtime/node_modules/aws-sdk/lib/sequential_executor.js:116:18)"
    ]
}
```

[...]

PR #711, opened 2024-03-04
fix(static-website): fix an issue in web acl creation with random WAFUnavailableEntityException

jessieweiyi:
Fixes #
#508 

Fix an issue where web acl failed creation when the IPSet is created but not querable yet ( intermittent WAFUnavailableEntityException).


cogwirrel (approved):

cogwirrel:
Looks like you need to commit the updated snapshots as it includes the asset hash of the lambda function which has changed :)

PR #712, opened 2024-03-04
feat(type-safe-api): expose managed rule props to allow consumers to …

mirgj:
Fixes # NA

This PR introduce the ability to modify the Waf managed rule set provided by the consumers in order to allow customisations that include, for example, the ability to change the individual rule behaviour inside a ruleset.

A common example is related to the `AWSManagedRulesCommonRuleSet` which has a strict rule named `SizeRestrictions_BODY` that allow only 8Kb body payload. With the current implementation provided by PDK the only option we have is to either disable Waf altogether or add a different rule set to override the default behaviour. The interface provided by type safe api doesn't provide the ability to modify the ruleset rules.

With this change, consumers can override individual rules in the ruleset as the entire CDK interface is shared to the consumer.

In order to have backwards compatibility this PR keep the usage of both `vendor` and `vendorName` properties. In future PDK iterations it's suggested to remove `vendor` to keep only `vendorName` using the default interface provided by CDK


packages/type-safe-api/src/construct/waf/open-api-gateway-web-acl.ts
```diff
@@ -46,8 +46,20 @@ export class OpenApiGatewayWebAcl extends Construct {
         })
       : undefined;
 
+    // TODO: vendor property is deprecated, to be removed in the future iterations
+    // and vendorName will be required
+    const anyMissingVendor = props.managedRules?.some(
+      (q) => !q.vendorName && !q.vendor
+    );
+
+    if (anyMissingVendor === true) {
```
cogwirrel: Nit: can simplify to `if (anyMissingVendor) {` :)

packages/type-safe-api/src/construct/waf/types.ts
```diff
@@ -1,15 +1,32 @@
...
+ * The type ManagedRule has been used to guarantee backward compatibility
+ * with forward compatibility provided by the vendorName property
+ *
+ * in future releases vendorName should be required while vendor will be removed
+ */
+export interface ManagedRule extends PartialManagedRuleGroupStatementProperty {
   /**
    * The name of the managed rule group vendor. You use this, along with the rule group name, to identify the rule group.
+   *
+   * @deprecated use the `vendorName` property instead. This property will be removed in the next releases.
```
cogwirrel: nit: `removed in the next major release`

cogwirrel (approved):
Looks great, thank you! Just a couple of minor nit-picks :)

PR #720, opened 2024-03-13
fix: ensure python poetry envs are created when installing

agdimech:
- use pnpm instead of yarn for non-node projects.
- force usage of pyenv as a global dependency as this leads to more predicatable installs/builds and less errors to do with python aliases, etc 
- Ensure all non-node projects have their `install:ci` target called as part of the root `install:ci` script
- Ensure all python projects call `poetry env use pythonXXX` before installing to ensure the venv is setup correctly.
  - The logic here is to parse the `pyproject.toml` and fetch the listed `python` version. If it starts with a '^', we fetch the latest installed python version from the user's machine matching the major version. If the version is pinned, then we use that version.
 - Update docs to mention pyenv should be used.
 - Replace yarn with pnpm in getting started.


packages/cloudscape-react-ts-website/src/cloudscape-react-ts-website-project.ts
```diff
@@ -59,7 +59,7 @@ export class CloudscapeReactTsWebsiteProject extends ReactTypeScriptProject {
       packageManager:
         options.parent && options.parent instanceof NodeProject
           ? options.parent.package.packageManager
-          : options.packageManager,
+          : NodePackageManager.PNPM,
```
cogwirrel: Should this be `: (options.packageManager ?? NodePackageManager.PNPM)` to respect the user specified one if provided?

packages/monorepo/src/components/nx-configurator.ts
```diff
@@ -383,18 +401,56 @@ export class NxConfigurator extends Component implements INxProjectCore {
...
+            .filter((t) => t.name !== "install")
+            .reduce(
+              (p, c) => ({
+                ...p,
+                [c.name]: NodePackageUtils.command.projen(
+                  NodePackageManager.PNPM,
+                  c.name
+                ),
+              }),
+              {}
```
cogwirrel: Nit: prefer `Object.fromEntries(this.project.tasks.all.filter(...).map((c) => [c.name, NodePackageUtils.command...]))` :)

packages/monorepo/src/components/nx-configurator.ts
```diff
@@ -383,18 +401,56 @@ export class NxConfigurator extends Component implements INxProjectCore {
...
       }).synthesize();
     }
   }
 
+  private _invokeInstallCITasks() {
+    const cmd = NodePackageUtils.command.exec(
+      ProjectUtils.isNamedInstanceOf(this.project, NodeProject)
+        ? this.project.package.packageManager
+        : NodePackageManager.NPM,
+      "nx run-many -t install:ci--output-style=stream --nx-bail"
```
cogwirrel: Missing a space here in `install:ci--output-style=stream`? :) Also I think there's a helper method for creating the run-many commands iirc :)

packages/monorepo/src/components/nx-configurator.ts
```diff
@@ -383,18 +401,56 @@ export class NxConfigurator extends Component implements INxProjectCore {
...
+    );
+    const task = this.project.tasks.tryFind("install:ci");
+    task?.steps[task.steps.length - 1]?.exec !== cmd && task?.exec(cmd);
```
cogwirrel: I think you'll need to also do `task?.steps?.length` to avoid accessing a property of undefined here :)

cogwirrel (commented):
Nice, this is going to make python a lot more robust :)

packages/cloudscape-react-ts-website/src/cloudscape-react-ts-website-project.ts
agdimech: done!

packages/monorepo/src/components/nx-configurator.ts
agdimech: Done.

packages/monorepo/src/components/nx-configurator.ts
agdimech: fixed.

packages/monorepo/src/components/nx-configurator.ts
agdimech: Fixed.

cogwirrel (approved):

PR #727, opened 2024-03-19
feat: pin java dependency versions

cogwirrel:
Pin all Java dependency versions, including PDK (using read-pkg-up to set the PDK dependency version to the one that synth'd the pom.xml)


agdimech (approved):
Looks good, let's add the version upgrade plugin as a next step.

PR #732, opened 2024-03-21
feat: generate openapitools.json via task

zsstiers:
Create a task that produces the openapitools.json file. This allows consumers to add their own steps to edit the file with compile-time values prior to being consumed. This is important for cases such as inserting a signed repository download URL.


cogwirrel (approved):
Nice work, thank you! 😄

agdimech (approved):

PR #738, opened 2024-03-28
fix(type-safe-api): fix websocket api issues

cogwirrel:
Fixes a few teething issues with the new websocket api:

* The async transformer jar wasn't included in the cache output, and so wasn't published in the release build
* Missed substituting the namespace for some of the async Smithy validators
* A couple of issues with the getting started guide projenrc sample


martenpayne (approved):
all good 👍

Issue #736, opened 2024-03-25
[BUG] Projects which set licensed to false emitting Apache 2.0


zsstiers:
### Describe the bug

When upgrading to the latest version of PDK unexpected license files have appeared.

### Expected Behavior

No license to be generated when a subproject sets `licensed: false`.

### Current Behavior

Many LICENSE files have been generated even though their projects set `licensed: false`.

### Reproduction Steps

```
import { javascript } from "projen";
import { monorepo } from "@aws/pdk";
import { NodeProject } from "projen/lib/javascript";

const project = new monorepo.MonorepoTsProject({
  devDeps: ["@aws/pdk"],
  name: "example",
  packageManager: javascript.NodePackageManager.PNPM,
  projenrcTs: true,
});

new NodeProject({
  parent: project,
  name: "nolicense",
  outdir: "packages/nolicense",
  defaultReleaseBranch: "main",
  licensed: false,
  packageManager: javascript.NodePackageManager.PNPM,
});

project.synth();
```
produces packages/nolicense/LICENSE as an Apache 2.0 license even though it specifically set licensed to false.

### Possible Solution

Sadly I don't see a clear path to easily support this because the `licensed` property is not made available anywhere. Maybe just a way to opt-out of the `_addLicenses` feature entirely?

### PDK version used

0.23.19

### What languages are you seeing this issue on?

Typescript

### Environment details (OS name and version, etc.)

Amazon AL2

PR #746, opened 2024-03-31
fix: no LICENSE file if licensed param is false

sebastianrothbucher:
Fixes the issue that `licensed: false` still generates a LICENSE file, which is at least confusing for internal code


sebastianrothbucher:
seems we had the same idea at the same time - #736 does just what I wanted to do. We can forget about mine

PR #796, opened 2024-06-18
feat: add support for enabling self signup in Cognito

agdimech:
Add support for enabling users to opt-in to self registration on the Cognito User Pool. This PR makes the following changes:

- Add a allowSignup option to be passing into the CloudscapeReactTSWebsite so that the UI can render the signup page.
- Add a allowSignup option to the infrastructure packages to enable signup on the Cognito User Pool.
- Modify Infrastructure templates to enable signup in the generated application-stack.


martenpayne (approved):
looks good

PR #800, opened 2024-06-24
feat(infra): add support for stages in infra sample code generator

agdimech:
Provide the ability to configure multiple stacks in the generated `main` infra based on stage configuration.


martenpayne (approved):
looks good

PR #801, opened 2024-06-27
fix: upgrade deps

agdimech:
- Upgrades all PDK deps to latest
- Ensures all tasks are rendered correctly, regardless of whether projen is ejected or not.
- Fixes Dependabot issues
- Changed websocket unit test to only fail if errors are thrown (currently will fail if warnings are found)
- Fixed issue in cdk-graph where return type should be boolean which was resulting in ts errors
- Fixed a bug in the Python infra sample which was causing a synth error


martenpayne (approved):
looks good

PR #816, opened 2024-07-24
feat(type-safe-api): allow customized output spec bucket (#812)

### LINKED ISSUE #812: [FEATURE] Add ability to customize outputSpecLocation S3 bucket for type-safe-rest-api PrepareApiSpecCustomResource
### Describe the feature

Allow customization of the outputSpecLocation in type-safe-rest-api, so the PrepareApiSpecCustomResource writes to the provided bucket instead of the assets bucket.

### Use Case

I am trying to deploy a type-safe-api using an environment that has read-only access to the assets s3 bucket. This deployment fails because type-safe-api tries to write the prepared specification to the assets bucket, and there's currently no good way to work around it. 

### Proposed Solution

Expose the outputSpecLocation as a customizable property for consumers of type-safe-rest-api. Otherwise, expose a boolean that handles the creation of a dedicated output bucket on the consumer's behalf.

### Other Information

_No response_

### Acknowledgements

### PDK version used

0.23.30

Typescript

MacOS

pcozzi:
Fixes #812 

This was manually tested by pulling it in to a monorepo and deploying. The prepared specification ended up in the provided bucket.

### DIFF
--- a/packages/type-safe-api/src/construct/type-safe-rest-api.ts
+++ b/packages/type-safe-api/src/construct/type-safe-rest-api.ts
@@ -27,6 +27,7 @@ import {
   Runtime,
 } from "aws-cdk-lib/aws-lambda";
 import { LogGroup } from "aws-cdk-lib/aws-logs";
+import { IBucket } from "aws-cdk-lib/aws-s3";
 import { Asset } from "aws-cdk-lib/aws-s3-assets";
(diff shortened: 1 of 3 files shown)


agdimech (approved):

agdimech:
Thanks for your contribution!

PR #831, opened 2024-09-30
feat(type-safe-api): replace openapi-generator for typescript rest api code generation

### LINKED ISSUE #789: [BUG] type-safe-api appending "operation" to some operationIds
### Describe the bug

For some of the operations with Python Lambda handlers defined in my OpenAPI spec, the operationId gets changed and has "operation" appended. When this happens, a handler file is created without the extra string, but the CDK infra has it in the handler path string.

### Expected Behavior

All of the operations defined in the spec should be named consistently, and the infrastructure should be consistent with module names.

### Current Behavior

Given a main.yaml including the following:
```yaml
  /generateSpec:
    post:
      operationId: generateSpec
      x-handler:
        language: python
      requestBody:
...
  /listSessions:
    get:
      description: List all sessions started in a time range.
      operationId: listSessions
      x-handler:
        language: python
      requestBody:
...
```

generates packages/api/generated/infrastructure/typescript/src/functions.ts
```typescript
/**
 * Options for the GenerateSpecFunction construct
 */
export interface GenerateSpecFunctionProps extends Omit<FunctionProps, 'code' | 'handler' | 'runtime'> {}

/**
 * Lambda function construct which points to the python implementation of GenerateSpec
 */
export class GenerateSpecFunction extends Function {
  constructor(scope: Construct, id: string, props?: GenerateSpecFunctionProps) {
    super(scope, id, {
      runtime: Runtime.PYTHON_3_12,
      handler: "api_detective_api_python_handlers.generate_spec.handler",
      code: Code.fromAsset(path.resolve(__dirname, "..",
        "../../../handlers/python/dist/lambda",
      )),
      tracing: Tracing.ACTIVE,
      timeout: Duration.seconds(30),
      ...props,
    });
  }
}
...
/**
 * Options for the ListSessionsOperationFunction construct
 */
export interface ListSessionsOperationFunctionProps extends Omit<FunctionProps, 'code' | 'handler' | 'runtime'> {}

/**
 * Lambda function construct which points to the python implementation of ListSessionsOperation
 */
export class ListSessionsOperationFunction extends Function {
  constructor(scope: Construct, id: string, props?: ListSessionsOperationFunctionProps) {
    super(scope, id, {
      runtime: Runtime.PYTHON_3_12,
      handler: "api_detective_api_python_handlers.list_sessions_operation.handler",
      code: Code.fromAsset(path.resolve(__dirname, "..",
        "../../../handlers/python/dist/lambda",
      )),
      tracing: Tracing.ACTIVE,
      timeout: Duration.seconds(30),
      ...props,
    });
  }
}
```

The handler file created for listSessions is inconsistent with the handler property.
packages/api/handlers/python/api_detective_api_python_handlers/generate_spec.py
packages/api/handlers/python/api_detective_api_python_handlers/list_sessions.py

### Reproduction Steps

I have no idea what's triggering this.

### Possible Solution

_No response_

I've created two new resources on this API in the last week. Both had this weird behavior. The first was named `commentSpec`. It was able to generate everything without "operation" appended when I changed the name to `iterateSpec`, but only after I deleted the nx cache, the packages/api directory, ran pdk, and restored the non-generated files.

### PDK version used

0.23.38 and 0.23.40

Typescript, Python

MacOS 14.4.1 intel

cogwirrel:
This change replaces the java openapi-generator CLI tool with our own code generation tool, which produces equivalent code. This removes the dependency on Java for the TypeScript REST API projects, as well as improving the speed of code generation by avoiding package installation steps and the multiple passes of openapi codegen used to work around limitations of the tool.

Note that this change fixes #789, which means that those hitting the edge case described in the issue may experience a breaking change.


packages/type-safe-api/scripts/type-safe-api/generators/typescript-cdk-infrastructure/templates/api.ejs
```diff
@@ -1,6 +1,14 @@
...
+  "id": "api",
+  "dir": "<%= metadata.srcDir || 'src' %>",
+  "name": "api",
+  "ext": ".ts",
+  "overwrite": true
+}
+###/TSAPI_WRITE_FILE###import { TypeSafeRestApi, TypeSafeRestApiProps, TypeSafeApiIntegration } from "@aws/pdk/type-safe-api";
 import { Construct } from "constructs";
-import { OperationLookup, OperationConfig } from "{{#apiInfo}}{{#apis.0}}{{vendorExtensions.x-runtime-package-name}}{{/apis.0}}{{/apiInfo}}";
+import { OperationLookup, OperationConfig } from "<%- metadata.runtimePackageName %>";
```
agdimech, line 11: EJS is so much cleaner/readable/better.

packages/type-safe-api/src/project/codegen/runtime/generated-typescript-async-runtime-project.ts
```diff
@@ -16,18 +27,74 @@ export interface GeneratedTypescriptAsyncRuntimeProjectOptions
...
+
+  protected readonly openapiGeneratorIgnore: OpenApiGeneratorIgnoreFile;
+
   constructor(options: GeneratedTypescriptAsyncRuntimeProjectOptions) {
     super(options);
 
     this.addDeps("@aws-sdk/client-apigatewaymanagementapi");
 
-    this.openapiGeneratorIgnore.addPatterns(
+    // TODO: remove this in favour of new codegen
```
agdimech, line 48: Any reason we can't do this now?

agdimech (approved):

packages/type-safe-api/src/project/codegen/runtime/generated-typescript-async-runtime-project.ts
cogwirrel: I just haven't written the ejs templates yet - will do as a follow up PR :)

PR #842, opened 2024-10-07
fix(monorepo): exclude projen from version update during monorepo upgrade-deps

### LINKED ISSUE #837: [BUG] MonorepoTsProject uses incorrect version of projen during "upgrade-deps"
### Describe the bug

This seems to occur when the latest published version of projen is a higher than the one hard coded in `projen-version.ts` .This can result (and actually does so for projen 0.88.0 vs 0.82.8) in changes to the tree from a transient use of the newer projen version, but then immediately reverting projen to the hard coded version, preventing the ability to adjust options to account for the changes.

### Expected Behavior

I expect that upgrade-deps does not make changes to the source tree that result from a transient use of a newer version of projen.

### Current Behavior

Running `upgrade-deps` results in side effects when it executes the `npm-check-updates` step.

### Reproduction Steps

1) Create a new monorepo (or us an existing one) using 0.23.54
2) Notice that out-of-the box, .gitatributes does not contain the default `text=auto eol=lf` introduced in projen 0.88.0
3) Run the `upgrade-deps` task
4) Notice that side effects of projen 0.88.0 appear in the .gitatributes files.

### Possible Solution

Consider adding `-x projen` to the command line for `npm-check-updates`.

The task for `upgrade-deps` includes a step running `npm-check-updates` on the entire monorep. This results in the projen version in the top/monorepo level package.json getting updated to the latest release.

A later step runs `pnpm i` (in the case of using pnpm), which installs the latest projen. The final step runs `pnpm exec projen` which:

- Executes the new version of projen, resulting in changes introduced by the new version of projen
- However, as projen runs, it eventually executes the MonorepoTsProject which forces reverting the projen version back to the one hard coded in `projen-version.ts`.
- Because the projen version was reverted, the projen base classes used by MonorepoTsProject do not support new options for TypeScriptProject (or its super classes) needed to control/counter the side effects.

### PDK version used

0.23.54 (possibly as early as 0.23.1)

Typescript

Ubuntu 22.04

memorsolutions:
Added `--reject projen` to `npm-check-updates` step of `upgrade-deps` task.

Fixes: #837

### DIFF
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -205,7 +205,7 @@
           "receiveArgs": true
         },
         {
-          "exec": "pnpm exec npm-check-updates --deep --rejectVersion 0.0.0 -u --dep prod,dev,peer,optional,bundle --target=minor"
+          "exec": "pnpm exec npm-check-updates --reject projen --deep --rejectVersion 0.0.0 -u --dep prod,dev,peer,optional,bundle --target=minor"
         },
         {
           "exec": "pnpm exec syncpack fix-mismatches"
--- a/packages/monorepo/src/projects/typescript/monorepo-ts.ts
+++ b/packages/monorepo/src/projects/typescript/monorepo-ts.ts
@@ -509,6 +509,7 @@ export class MonorepoTsProject
         NodePackageUtils.command.exec(
           this.package.packageManager,
           "npm-check-updates",
+          "--reject projen",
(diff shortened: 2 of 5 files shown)


cogwirrel (approved):

cogwirrel:
Thanks so much for this! :)

PR #854, opened 2024-10-10
feat(cdk-graph-plugin-diagram): support predetermined node positions

### LINKED ISSUE #853: [FEATURE] Predetermined node positions
### Describe the feature

DOT supports the pos attribute (https://github.com/ts-graphviz/ts-graphviz/blob/af5a52474ca13065385b775d52d410b448d8b3a4/packages/common/src/attribute.ts#L2208-L2227) to define the position of a node. The `@aws/pdk/cdk-graph` (and the diagram plugin) should accept and account for a map of node ids and their positions (`Map<NodeId, Pos>`) to specify the location of specific nodes. This could apply for the DrawIO and the Graphviz provider.

### Use Case

While the current state of exported architecture diagrams is neat as a baseline, it often requires architects to make it easier to interpret by a humans. Manually editing the diagram to a certain layout is tedious and redundant if it can be automated by providing node positions given by ids, so it remains consistent between new diagram revisions.

### Proposed Solution

1. Extend the diagram config with

```typescript
export interface IDiagramConfigBase {
  /**
   * The output format(s) to generated.
   * @default `DiagramFormat.PNG` - which will through extension also generate `DiagramFormat.SVG` and `DiagramFormat.DOT`
   */
  readonly format?: DiagramFormat | DiagramFormat[];

  /** Graph {@link IGraphFilterPlan Filter Plan}  used to generate a unique diagram */
  readonly filterPlan?: IGraphFilterPlan;

  /** Config for graph theme */
  readonly theme?: GraphThemeConfigProp;

  /** Config for predetermined node positions */
  readonly positions?: Map<string, Point>;

}
```

2. extend https://github.com/aws/aws-pdk/blob/mainline/packages/cdk-graph/src/core/graph.ts#L1109 with a position field
3. and use those positions in https://github.com/aws/aws-pdk/blob/mainline/packages/cdk-graph-plugin-diagram/src/internal/graphviz/entities/nodes.ts#L66 to set the `pos` attribute in the emitted DOT output.

### Other Information

_No response_

### Acknowledgements

### PDK version used

0.23.37

Typescript

MacOS Sequoia 15.0.1

martinjlowm:
Fixes #853 by adding a `nodePositions` configuration option that takes a `Map` of construct ids and maps them to fixed positions.

One caveat is that the default dot-layout engine ignores `pos` attributes, so we have to resort to a different one. `fdp` seems like the most reliable in the context of diagrams. The engine only changes if `nodePositions` is provided and it isn't empty.

Setting a fixed location for WebServer (`(0, 10)`) results in the following diagram.

diagram


packages/cdk-graph-plugin-diagram/src/config.ts
```diff
@@ -51,6 +52,9 @@ export interface IDiagramConfigBase {
 
   /** Config for graph theme */
   readonly theme?: GraphThemeConfigProp;
+
+  /** Config for predetermined node positions */
+  readonly nodePositions?: Map<string, Point>;
```
cogwirrel: Looks like the build is failing because JSII doesn't like `Map` - I think you should be able to do an object type like: `{ [nodeId: string]: Point }`

Also I think we might need to define our own `Point` interface as I don't think ts-graphviz uses JSII :)

packages/cdk-graph-plugin-diagram/src/config.ts
```diff
@@ -51,6 +52,9 @@ export interface IDiagramConfigBase {
 
   /** Config for graph theme */
   readonly theme?: GraphThemeConfigProp;
+
+  /** Config for predetermined node positions */
```
cogwirrel: Would be good to mention that the key is a Node ID / CDK construct ID in the comment :)

packages/cdk-graph-plugin-diagram/test/graphviz/dot.test.ts
```diff
@@ -189,4 +189,32 @@ describe("dot", () => {
...
     });
   });
+
+  it("should set node position", async () => {
+    const outdir = await makeCdkOutdir("node-position");
+
+    const app = new FixtureApp({ outdir });
+    const plugin = new CdkGraphDiagramPlugin({
+      defaults: {
+        nodePositions: new Map([["WebServer", "0,10!"]]),
```
cogwirrel: Hmm is `Point` just a string? We'd need to make sure we document some examples if we leave it as a string - I'd have expected an object with x/y properties :)

cogwirrel (commented):
That was quick! Thanks so much! 🤩

Just a couple of minor comments :)

Would be great if you could add an example of how to use it in the docs too :)

martinjlowm:
Now accounting for your comments - let me know what you think. Happy to elaborate on the docs (had to make it short as I'm on my way to the office.

@cogwirrel

packages/cdk-graph-plugin-diagram/src/internal/graphviz/entities/nodes.ts
```diff
@@ -59,6 +59,11 @@ function marshallLabelForRendering(original: string): MarshalledLabel {
   return { original, label, lines: lines.length };
 }
 
+/**
+ * Positional coordinates for a node
+ */
+export type NodePosition = { x: number; y: number };
```
cogwirrel: Please could we specify what the units are for these? I think maybe it's inches for graphviz?

cogwirrel:
> Now accounting for your comments - let me know what you think. Happy to elaborate on the docs (had to make it short as I'm on my way to the office.
> 
> @cogwirrel

This looks great to me! Just one minor comment :)

You might also need to specify `{ [cdkConstructId: string]: NodePosition }` for JSII to be happy :) You may also need to change `NodePosition` into an `interface` rather than a `type` for JSII too but I'm not sure - it's worth building locally and iterating if you're able :)

packages/cdk-graph-plugin-diagram/src/internal/graphviz/entities/nodes.ts
martinjlowm: Inches it is indeed

cogwirrel (approved):

cogwirrel:
Ooh looks like you need to make sure that `index.ts` exports `NodePositions` :)

martinjlowm:
Let's see if this works ✌️

cogwirrel:
Ah sorry I should have been more specific! You'll need to export from `packages/cdk-graph-plugin-diagram/src/index.ts` :) I feel like we're really close though! 😄

martinjlowm:
That was a bit of trial and error. Turns out I had to move out the interface from the internal directory. Apparently, jsii considered it internal similarly to commented interfaces. I also prefixed it with `I` to be interpreted as a behavioral interface. I'm pretty new to the jsii type system, so if you think it would be better to consider it as a struct, I'd be happy to revert it.

@cogwirrel

cogwirrel (approved):

cogwirrel:
@martinjlowm Ah great find! Thanks so much for this! An interface seems like it's consistent with the rest of the configuration so sounds great to me! 😄

PR #867, opened 2024-10-21
feat(type-safe-api): faster code generation for java runtime

cogwirrel:
Move java runtime to new codegen.

Note that we no longer generate markdown docs as part of the Java runtime, but users can generate similar markdown docs in the dedicated documentation project.


packages/type-safe-api/scripts/type-safe-api/generators/generate-next.ts
```diff
@@ -178,6 +178,40 @@ const toPythonName = (namedEntity: 'model' | 'property' | 'operation', name: str
   return nameSnakeCase;
 };
```
martenpayne, line 180: is the list below complete against https://en.wikipedia.org/wiki/List_of_Java_keywords ?

packages/type-safe-api/scripts/type-safe-api/generators/java/templates/client/auth/apiKeyAuth.ejs
```diff
@@ -0,0 +1,76 @@
...
+  "overwrite": true
+}
+###/TSAPI_WRITE_FILE###<%- include('../../header.partial.ejs', { info }) %>
+
+
+package <%- metadata.packageName %>.auth;
+
+import <%- metadata.packageName %>.ApiException;
+import <%- metadata.packageName %>.Pair;
+
```
martenpayne, line 15: can the bulk of these Java module templates be read in as a data resource at runtime (possibly allowing for some level of validation with javac)?

packages/type-safe-api/scripts/type-safe-api/generators/java/templates/server/operationConfig.ejs
```diff
@@ -0,0 +1,30 @@
...
+
+// Generic type for object "keyed" by operation names
+@lombok.Builder @lombok.Getter
+public class OperationConfig<T> {
+    <%_ allOperations.forEach((operation) => { _%>
+    private T <%- operation.name %>;
+    <%_ }); _%>
+
+    public Map<String, T> asMap() {
+        Map<String, T> map = new HashMap<>();
```
martenpayne, line 24: I'm a little rusty on Java generics; is this <> syntax valid?

martenpayne (commented):
Just left a couple questions/comments…

packages/type-safe-api/scripts/type-safe-api/generators/generate-next.ts
cogwirrel: It's the same as OpenAPI generator's here for parity :) https://github.com/OpenAPITools/openapi-generator/blob/8f2676c5c2bcbcc41942307e5c8648cee38bcc44/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractJavaCodegen.java#L179

I'll add a comment to the code to link it!

packages/type-safe-api/scripts/type-safe-api/generators/java/templates/server/operationConfig.ejs
cogwirrel: [Yes!](https://docs.oracle.com/javase/tutorial/java/generics/types.html?ref=sangkon.com#:~:text=In%20Java%20SE%207%20and,type%20arguments%20from%20the%20context.) 😄

packages/type-safe-api/scripts/type-safe-api/generators/java/templates/client/auth/apiKeyAuth.ejs
cogwirrel: I think compiling the generated code as part of some kind of integration test would be the way to go here - I've raised https://github.com/aws/aws-pdk/issues/868 to track that :)

martenpayne (approved):

PR #872, opened 2024-10-24
feat(type-safe-api): separate model projects per model language

cogwirrel:
Rather than sharing a single `TypeSafeApiModelProject` for both Smithy and OpenAPI, we separate them into their own projects. This improves the ability to add future additional model languages since they can inherit from other projen project types (not just `Project`). This also makes instantiating standalone Smithy libraries more intuitive as they are created with `SmithyModelProject`, rather than `TypeSafeApiModelProject` and modelLanguage set to SMITHY, therefore no `!` needs to be used to access the possibly undefined `smithy` property.

BREAKING CHANGE: Removed `TypeSafeApiModelProject`. Please instantiate Smithy shape libraries with `SmithyModelProject`.

re #825 - this change will also make it easier to add the functionality requested for a smithy shape library with no openapi projection / build step.

NB: this does not change the project structure in any way, it's purely a refactor of the projen constructs and therefore there are no jest snapshot changes.


agdimech (approved):
Looks good!

PR #874, opened 2024-10-25
feat(type-safe-api): support for modelling apis in typespec

cogwirrel:
Adds support for modelling both REST and WebSocket APIs in TypeSpec: https://typespec.io/

Similar to Smithy, TypeSpec provides a nice abstraction for modelling API operations and their inputs and outputs. TypeSpec syntax takes some inspiration from TypeScript, and so may be more intuitive for some users. TypeSpec operates in the Node ecosystem, and so no additional global dependencies are required for its use (unlike Smithy which requires Java and Maven).


agdimech (approved):
Noice

PR #882, opened 2024-11-08
fix(type-safe-api): fix model serialisation for nested collections in typescript

cogwirrel:
Previously, arrays of arrays or dictionaries of dictionaries (or any combination) were not serialising/deserialising correctly. There were 2 main problems:

1. Nested arrays/dictionaries of models would be treated as if they were a single array/dictionary
2. Nested dates (even under single level arrays) weren't being serialised/deserialised at all which could lead to runtime errors

Note that OpenAPI generator would not attempt to recursively serialise/deserialise nested arrays which mean that problem 2 was present prior to moving to the new code generator.


agdimech (approved):
Lgtm

PR #926, opened 2025-06-16
feat(type-safe-api): add support for additional request parameters in s3 integration

v41e:
This PR adds logic to support additional request parameters and mapping to path, improving integration with S3.

Changes:
- added handling for extra request parameters in the construct props
- implemented mapping logic to correctly resolve and map paths for S3 integration

### DIFF
--- a/packages/type-safe-api/src/construct/integrations/s3.ts
+++ b/packages/type-safe-api/src/construct/integrations/s3.ts
@@ -43,14 +43,12 @@ export interface S3IntegrationProps {
 
   /**
    * Specifies additional query string request parameters to be passed to the integration request.
-   * These parameters will be mapped to path parameters in order to allow for dynamic S3 bucket paths.
    * @default - no additional query string request parameters
    */
   readonly queryStringRequestParameters?: string[];
 
   /**
    * Specifies additional header request parameters to be passed to the integration request.
-   * These parameters will be mapped to path parameters in order to allow for dynamic S3 bucket paths.
    *  @default - no additional header request parameters
    */
   readonly headerRequestParameters?: string[];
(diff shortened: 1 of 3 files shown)


packages/type-safe-api/src/construct/integrations/s3.ts
```diff
@@ -40,6 +40,20 @@ export interface S3IntegrationProps {
...
+   * @default - no additional query string request parameters
+   */
+  readonly queryStringRequestParameters?: string[];
+
+  /**
+   * Specifies additional header request parameters to be passed to the integration request.
+   * These parameters will be mapped to path parameters in order to allow for dynamic S3 bucket paths.
+   *  @default - no additional header request parameters
+   */
+  readonly headerRequestParameters?: string[];
```
cogwirrel: Nit: let's remove `These parameters will be mapped to path parameters in order to allow for dynamic S3 bucket paths.` :)

cogwirrel (commented):
Looks great, thank you! 😄 

Please could you add a simple unit test for these changes too?

v41e:
Done, I hope AWS will move from JSONpath to JSONata in API Gateway (like they did in Step Function) in order to allow simple transformation.

For example to allow to set a default value if a query param is missing.

cogwirrel:
> Done, I hope AWS will move from JSONpath to JSONata in API Gateway (like they did in Step Function) in order to allow simple transformation.
> 
> For example to allow to set a default value if a query param is missing.

Thank you! :)

I wonder if you can achieve that with APIGateway's VTL, eg something like this? https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-override-request-response-parameters.html#apigateway-override-request-response-examples-2

cogwirrel (approved):

v41e:
i didn't know request template has a `context.requestOverride`

PR #938, opened 2026-01-30
feat(deps): upgrade default runtimes, projen version, clean deprecated and stabilize workspace tooling

v41e:
## Summary

- Upgrade default Java, Python, and Node runtimes to the latest stable versions.
- Stabilize PNPM + jsii-docgen behavior in the monorepo.
- Add a license allowlist workaround for an indirect dependency.

Review note: I split changes into logical, isolated commits to make review easier and allow selective feedback/reverts.

## Context / Why

I bumped default runtimes to keep the PDK template current. During the upgrade I hit a few tooling issues (jsii-docgen resolution in PNPM layouts and license auditing). This PR includes pragmatic workarounds to keep builds green while we discuss better long‑term options.

## What’s changed

- **Runtime defaults:** Java, Python, and Node versions bumped to latest stable.
- **Workspace tooling:** set `node-linker=hoisted` to avoid PNPM resolution issues affecting jsii-docgen/Nx tools.
- **Docgen stability:** jsii-docgen resolution improvement (monorepo).
- **License check:** added `Python-2.0` to allowlist due to `argparse` via `js-yaml`.

## Impact / Notes
- CI/tooling images pinned to old runtime versions may need updating.
- Hoisted node_modules is a compatibility tradeoff; it can relax dependency boundaries.
- The Python‑2.0 allowlist entry is a workaround; ideally removed once upstream deps change.


cogwirrel:
Hey @v41e ! Thanks for your contribution, appreciate you breaking this down into multiple commits!

From a look through everything here seems reasonable. `Python-2.0` is ok to remain as an allow-listed license.

Are you able to look at updating the relevant github workflows for this to get the build passing? (Though looks like the build is currently failing due to an out of date pnpm-lock.yaml)

Also, are you able to test this end to end by creating, building and deploying a monorepo which uses each of the projects affected?

Please Note:

Our efforts are currently in the [Nx Plugin for AWS](https://github.com/awslabs/nx-plugin-for-aws) and we don't have the bandwidth to maintain PDK. We're happy to accept community contributions like this to keep it current, however we will most likely be deprecating PDK this year.

I'd recommend checking out the Nx Plugin for AWS to see if it fits your needs - it aims to solve the same problem (rapidly scaffold full-stack projects on AWS) but evolves on our learnings from PDK, aiming to reduce the learning curve and the amount of "magic". We have published a migration guide and FAQ [here](https://awslabs.github.io/nx-plugin-for-aws/en/blog/migrating-from-aws-pdk/) which may also be useful.

v41e:
Why you dropped projen in this new project?

cogwirrel:
> Why you dropped projen in this new project?

The primary reason is so that our tooling isn't a direct dependency. One of the great things about `projen` is that you can consume updates to keep your project structure/config up to date, however it relies upon the projects and components to be kept up to date. With the Nx Plugin for AWS we opted to generate once, then code/configuration is owned by consumers who are able to customise it as they see fit, which means we're not in the critical path. We've also found that using AI tools for development such as Kiro has been a bit more effective with the Nx Plugin for AWS as AI can update configuration in a normal way, and we opt to use more standard/well-known frameworks like tRPC and FastAPI.

Hope this helps!

